PR #4, opened 2017-04-15
Basic update phase functions registration and execution.

vadordev:


PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -2,34 +2,89 @@
...
 
-void Poly::Engine::Deinit()
+void Engine::Deinit()
 {
 	Renderer->Deinit();
 	delete Renderer;
 	Renderer = nullptr;
 }
 
+void Engine::RegisterPreupdatePhase(PhaseUpdateFunction phaseFunction)
```
MuniuDev: Let's pass these functions by reference to avoid double copy (1. on call, 2. on PushBack)

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -1,31 +1,72 @@
...
 		void Deinit();
 
+		//////////////////////////////
+		/// Registers a PhaseUpdateFunction to be executed in the Preupdate 
+		/// part of a single frame in the same order as they were passed in.
+		/// @see RegisterUpdatePhase()
+		/// @see RegisterPostupdatePhase()
+		/// @param phaseFunction - void function(World*)
+		//////////////////////////////
+		void RegisterPreupdatePhase(PhaseUpdateFunction phaseFunction);
```
MuniuDev: We can avoid code duplication (triplication?) by using enum class designating all three phases as an input argument.

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -1,31 +1,72 @@
...
+		/// @param phaseFunction - void function(World*)
+		//////////////////////////////
+		void RegisterPostupdatePhase(PhaseUpdateFunction phaseFunction);
+
+
 		void Update(float dt);
 
 		World& GetWorld() { return BaseWorld; }
 	private:
+		inline void PreupdatePhases();
```
MuniuDev: inline keywoard won't have any effect if you move implementation outside of the header file.

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -1,31 +1,72 @@
...
+		inline void UpdatePhases();
+		inline void PostupdatePhases();
+
 		World BaseWorld;
 		IGame* Game;
 		IRenderingContext* Renderer;
+		Dynarray< PhaseUpdateFunction > GamePreupdatePhases;
```
MuniuDev: enum class can be used here as well `Dynarray< PhaseUpdateFunction > GamePhases[static_cast<int>(eUpdateMode::_COUNT)]`. We will create EnumArray for such cases in the future.

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -28,32 +28,25 @@ namespace Poly
...
-		/// part of a single frame in the same order as they were passed in.
+		/// Enum used to specify the part of update to which a given phase is registered.
 		/// @see RegisterUpdatePhase()
-		/// @see RegisterPostupdatePhase()
-		/// @param phaseFunction - void function(World*)
 		//////////////////////////////
-		void RegisterPreupdatePhase(PhaseUpdateFunction phaseFunction);
+		enum class eUpdatePhaseOrder 
+		{
+			Preupdate,
```
MuniuDev: According to our coding style enum fields should by all capital letters with aunderscores, i.e. PRE_UPDATE, _COUNT.

PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -23,46 +23,28 @@ void Engine::Deinit()
...
 
-void Engine::RegisterPreupdatePhase(PhaseUpdateFunction phaseFunction)
+void Engine::RegisterUpdatePhase(const PhaseUpdateFunction& phaseFunction, eUpdatePhaseOrder order)
 {
-	for (auto& iter : GamePreupdatePhases)
+	HEAVY_ASSERTE(order != eUpdatePhaseOrder::_Count, "_Count enum value passed to RegisterUpdatePhase(), which is an invalid value");
+	Dynarray<PhaseUpdateFunction>& UpdatePhases = GameUpdatePhases[static_cast<int>(order)];
+	for (auto& iter : UpdatePhases)
 	{
 		if (iter.target<PhaseUpdateFunction>() == phaseFunction.target<PhaseUpdateFunction>())
```
MuniuDev: This should be an assertion. We don't want it to silently fail.

PR #6, opened 2017-04-16
Implemented GetSibling() for components

vadordev:
... along with a few component getters and sanity checkers.

### DIFF
--- a/PolyEngine/Engine/Src/Entity.hpp
+++ b/PolyEngine/Engine/Src/Entity.hpp
@@ -1,10 +1,12 @@
 #pragma once
 
 #include <Core.hpp>
+#include <bitset>
 
 #include "ComponentBase.hpp"
 
-namespace Poly {
+namespace Poly 
+{
 
 	class World;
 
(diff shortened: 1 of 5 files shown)


PolyEngine/Engine/Src/Entity.hpp
```diff
@@ -16,12 +18,48 @@ namespace Poly {
...
+		//////////////////////////////
+		/// Note: This function should only be accessed by engine's internal logic.
+		/// Checks whether there are all of the specified components under this Entity's ID.
+		/// @param IDs is a bit field of component IDs to be checked for, see macro BIT()
+		/// @see GetComponent()
+		/// @see HasComponent()
+		//////////////////////////////
+		inline bool HasComponents(unsigned long int IDs) const
+		{
+			return ComponentPosessionFlags.to_ullong() & IDs;
```
MuniuDev: This method will throw exception in size of the bit set is greater than 64bits. Let's leave it for now and add a comment stating that BitMask class impleentation is required here + a STATIC_ASSERTE for MAX_COMPONENTS_COUNT <= 64.

PR #8, opened 2017-04-25
InputPhase

PiotrMoscicki:
- registered InputPhase
- tests for WorldInputComponent in InputPhase


PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -0,0 +1,80 @@
...
+using namespace Poly;
+
+void InputSystem::InputPhase(World* world)
+{
+	InputWorldComponent* com = world->GetInputWorldComponent();
+
+	for (int i = 0; i < 256; i++)
+		com->PrevKey[i] = com->CurrKey[i];
+
+	//memcpy(&com->PrevKey, &com->CurrKey, 256 * sizeof(bool));
```
MuniuDev: This memcpy would be much faster.

PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "World.hpp"
+#include "InputWorldComponent.hpp"
+
+using namespace Poly;
+
+void InputSystem::InputPhase(World* world)
+{
+	InputWorldComponent* com = world->GetInputWorldComponent();
+
+	for (int i = 0; i < 256; i++)
```
MuniuDev: Let's make 256 a constant. It's used in many places.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -0,0 +1,33 @@
...
+			bool GetPressed(eKey key1, eKey key2)			 { return (CurrKey[static_cast<int>(key1)] && CurrKey[static_cast<int>(key2)]) ? true : false; }
+			bool GetPressed(eKey key1, eKey key2, eKey key3) { return (CurrKey[static_cast<int>(key1)] && CurrKey[static_cast<int>(key2)] && CurrKey[static_cast<int>(key3)]) ? true : false; }
+			bool GetClicked(eKey key)  { return (CurrKey[static_cast<int>(key)] && !PrevKey[static_cast<int>(key)]) ? true : false; }
+			bool GetReleased(eKey key) { return (!CurrKey[static_cast<int>(key)] && PrevKey[static_cast<int>(key)]) ? true : false; }
+			const Vector& GetMousePos() { return CurrMouse; }
+			const Vector& GetMouseDiff() { return CurrMouse - PrevMouse; }
+			const int& GetWheelPos() { return CurrWheel; }
+			const int& GetWheelDiff() { return CurrWheel - PrevWheel; }
+		private:
+			bool CurrKey[256];
```
MuniuDev: You can use EnumArray<bool, eKey> for this. Then those ugly static casts will be unnecessary.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace Poly
+{
+	namespace InputSystem
+	{
+		class InputWorldComponent : public ComponentBase
+		{
+		friend void InputPhase(World*);
+		public:
+			bool GetPressed(eKey key)						 { return CurrKey[static_cast<int>(key)]; }
+			bool GetPressed(eKey key1, eKey key2)			 { return (CurrKey[static_cast<int>(key1)] && CurrKey[static_cast<int>(key2)]) ? true : false; }
```
MuniuDev: If we want a method that checks multiple keys, lets write it with an argument std::initializer_list<eKey>.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <Vector.hpp>
+#include "ComponentBase.hpp"
+#include "KeyBindings.hpp"
+
+namespace Poly
+{
+	namespace InputSystem
+	{
+		class InputWorldComponent : public ComponentBase
```
MuniuDev: plase add ENGINE_DLLEXPORT

PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -0,0 +1,80 @@
...
+		else if(InputEventsQueue->Front().Type == eInputEventType::KEYUP)
+			gConsole.LogDebug("Keyup: {}", (unsigned)InputEventsQueue->Front().Key);
+		else if(InputEventsQueue->Front().Type == eInputEventType::MOUSEMOVE)
+			gConsole.LogDebug("Mousemove: {}", InputEventsQueue->Front().Pos);
+		else if(InputEventsQueue->Front().Type == eInputEventType::WHEELMOVE)
+			gConsole.LogDebug("Wheelmoove: {}", InputEventsQueue->Front().Pos);
+		InputEventsQueue->Pop();
+	}*/
+
+	if(com->GetPressed(eKey::ALT)) gConsole.LogDebug("ALT pressed!!");
```
MuniuDev: These explicit test methods should be removed before merge.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -0,0 +1,33 @@
...
+		{
+		friend void InputPhase(World*);
+		public:
+			bool GetPressed(eKey key)						 { return CurrKey[static_cast<int>(key)]; }
```
MuniuDev: All those getters should be const methods. Also consider renaming them from GetPressed to IsPressed, and similarily for other methods.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <Vector.hpp>
+#include "ComponentBase.hpp"
+#include "KeyBindings.hpp"
+
+namespace Poly
+{
+	namespace InputSystem
```
MuniuDev: Please don't put components in system namespaces.

PolyEngine/Engine/Src/World.hpp
```diff
@@ -80,6 +82,8 @@ namespace Poly {
 		}
 
 		Engine* GetEngine() const { return EnginePtr; }
+		InputSystem::InputWorldComponent* GetInputWorldComponent() { return &InputCom; };
```
MuniuDev: Return reference if possible. Here it's possible.

PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -0,0 +1,80 @@
...
+			com->CurrMouse = InputEventsQueue->Front().Pos;
+			break;
+		case eInputEventType::WHEELMOVE:
+			com->CurrWheel += InputEventsQueue->Front().Pos.X;
+			break;
+		}
+		InputEventsQueue->Pop();
+	}
+
+	/*while(InputEventsQueue->Size() > 0){
```
MuniuDev: Either remove this all together or put debug messeges in the preceding while statement.

PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -7,74 +7,35 @@ using namespace Poly;
 
 void InputSystem::InputPhase(World* world)
 {
-	InputWorldComponent* com = world->GetInputWorldComponent();
+	InputWorldComponent com = world->GetInputWorldComponent();
 
-	for (int i = 0; i < 256; i++)
-		com->PrevKey[i] = com->CurrKey[i];
+	memcpy(&com.PrevKey, &com.CurrKey, static_cast<int>(eKey::_COUNT) * sizeof(bool));
```
MuniuDev: You can use operator=() that is part of EnumArray. ;) Right now, this is undefined behaviour, you're relaying strictly on the fact that Data is the first member of EnumArray.

PolyEngine/Engine/Src/InputWorldComponent.hpp
```diff
@@ -1,33 +1,37 @@
...
 
 namespace Poly
 {
-	namespace InputSystem
+	class ENGINE_DLLEXPORT InputWorldComponent : public ComponentBase
 	{
-		class InputWorldComponent : public ComponentBase
+	friend void InputSystem::InputPhase(World*);
+	public:
+		const bool IsPressed(const std::initializer_list<eKey>& list)
```
MuniuDev: Alongside this method please add `bool IsPressed(eKey key)` as a convinience. Move the implementation to cpp.
And the const needs to be after method specification. for example: `bool IsPressed(const std::initializer_list<eKey>& list) const;`

PolyEngine/Engine/Src/KeyBindings.hpp
```diff
@@ -177,6 +177,7 @@ namespace Poly
 		PLAY =					VK_PLAY,
 		ZOOM =					VK_ZOOM,
 		PA1 =					VK_PA1,
-		OEM_CLEAR =				VK_OEM_CLEAR
+		OEM_CLEAR =				VK_OEM_CLEAR,
+		_COUNT
```
MuniuDev: Right now value of `_COUNT == VK_OEM_CLEAR + 1` is this the intended behaviour? Maybe set explicitly _COUNT to 256.

PR #10, opened 2017-04-27
Added Queue.

szymon112233:
Pls don't kill me :3

### DIFF
--- a/PolyEngine/Core/Src/Queue.hpp
+++ b/PolyEngine/Core/Src/Queue.hpp
@@ -9,17 +9,88 @@ namespace Poly
 	{
 	public:
 		Queue() = default;
-		Queue(size_t size) : Data(size) {};
+		Queue(size_t size) : Data(size), RingCap(size), BuffCap(size)
+		{
+			for (size_t i = 0; i < size; i++)
+			{
+				Data.PushBack(T());
+			}
+		}
 
-		bool Empty() const { return Data.IsEmpty(); }
-		size_t Size() const { return Data.GetSize(); }
-		const T& Front() const { return Data.GetData()[0]; }
-		const T& Back() const { return Data.GetData()[Data.GetSize() - 1]; };
-		T& Front() { return Data.GetData()[0]; }
-		T& Back() { return Data.GetData()[Data.GetSize() - 1]; };
-		void Push(const T &obj) { Data.PushBack(obj); }
-		void Pop() { Data.PopFront(); }
+		bool Empty() const { return BuffSize == 0; }
+		size_t Size() const { return BuffSize; }
+		const T& Front() const { return Data.GetData()[Head]; }
+		const T& Back() const 
+		{
+			if (RingCap == BuffCap)
+			{
+				if (Tail - 1 < 0)
+					return Data.GetData()[RingCap - 1];
+				else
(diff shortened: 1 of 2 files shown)


PolyEngine/Core/Src/Queue.hpp
```diff
@@ -9,17 +9,88 @@ namespace Poly
...
+					return Data.GetData()[Tail -1];
+			}
+			else
+				return Data.GetData()[BuffCap - 1];
+		}
+		void Push(const T &obj) 
+		{
+			if (RingSize == RingCap || RingCap != BuffCap)
+			{
+				Data.PushBack(obj);
```
MuniuDev: Ok, I'm wrong, this seems to work well.

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -9,17 +9,88 @@ namespace Poly
...
-		T& Front() { return Data.GetData()[0]; }
-		T& Back() { return Data.GetData()[Data.GetSize() - 1]; };
-		void Push(const T &obj) { Data.PushBack(obj); }
-		void Pop() { Data.PopFront(); }
+		bool Empty() const { return BuffSize == 0; }
+		size_t Size() const { return BuffSize; }
+		const T& Front() const { return Data.GetData()[Head]; }
+		const T& Back() const 
+		{
+			if (RingCap == BuffCap)
```
MuniuDev: You can move this to private method `T& BackImpl() const` and then call this method in both versions of Back

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -9,17 +9,88 @@ namespace Poly
...
+			}
+			else
+			{
+				Data.GetData()[Tail] = obj;
+				Tail = (Tail + 1) % RingCap;
+				RingSize++;
+				BuffSize++;
+			}
+		}
+		void Pop()
```
MuniuDev: What about making this queue work in both directions (PopFront/Back and PushFront/Back? )

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -9,17 +9,88 @@ namespace Poly
...
+			 
+		}
 	private:
 		Dynarray<T> Data;
+		int Head = 0;
+		int Tail = 0;
+		size_t RingCap = 0;
+		size_t BuffCap = 0;
+		size_t RingSize = 0;
+		size_t BuffSize = 0;
```
MuniuDev: Isn't buff size always equall Data.size() and BuffCap always equal Data.GetCapacity() ? If so, please substitute them.

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -9,17 +9,88 @@ namespace Poly
 	{
 	public:
 		Queue() = default;
-		Queue(size_t size) : Data(size) {};
+		Queue(size_t size) : Data(size), RingCap(size), BuffCap(size)
+		{
+			for (size_t i = 0; i < size; i++)
```
MuniuDev: Actually this constructor should be totaly empty.

MuniuDev:
Won't merge. Different approach is wip.

PR #11, opened 2017-04-27
TransformComponentTests

R4yRayR4y:


PolyEngine/UnitTests/Src/TransformComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+TEST_CASE("TransformComponent with no parent", "[TransformCompoent]") 
+{
+	TransformComponent tc;
+	Vector translationVector = Vector(1, 2, 3);
+	Vector scaleVector = Vector(4, 5, 6);
+	SECTION("Empty constructor")
+	{
+		for (int i = 0; i < 16; ++i)
+		{
+			REQUIRE(tc.GetLocalTransformationMatrix().Data[i] == (i % 5 == 0 ? 1 : 0));
```
MuniuDev: Just compare wit Matrix(), this by default constructs identity matrix.

PolyEngine/UnitTests/Src/TransformComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+
+	// basic transform setting
+	tc.SetLocalTranslation(translationVector);
+	tc.SetLocalScale(scaleVector);
+
+	SECTION("Local transformation")
+	{
+		Matrix tcl = tc.GetLocalTransformationMatrix();
```
MuniuDev: instead of comparing matrix fields it would be much better to simulate some real life sceniarios and compare the output witrh the prediction. For example: Create two related component and the apply transform to parent and chck if the child transfomr changed accordingly, same with rotation and scale.

You should definitely use GetGlobalRotation/Translation/Scale methods for this.

This comment applies to all unit tests in this file.

PolyEngine/UnitTests/Src/TransformComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+			Angle::FromDegrees(90) };
+		Quaternion rotation(angles);
+		tc2.SetLocalRotation(rotation);
+		float data[] = {0,-1,0,0,
+			  1,0,0,0,
+			  0,0,1,0,
+			  0,0,0,1};
+		for (int i = 0; i < 16; i++)
+		{
+			REQUIRE(tc2.GetLocalTransformationMatrix().Data[i] - data[i] < 0.000001);
```
MuniuDev: use function bool Cmpf(float, float), which compares two floats and return true if equall.

PolyEngine/UnitTests/Src/TransformComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+		float scale1 = 2;
+		float scale2 = 3;
+		parent.SetLocalScale(scale1);
+		child.SetLocalScale(scale2);
+		REQUIRE(child.GetGlobalScale() == Vector(6, 6, 6));
+	}
+
+	SECTION("Rotation")
+	{
+		EulerAngles angles1 = { Angle::FromDegrees(10),
```
MuniuDev: Fun fact. You can also use literals (10_deg, 20_deg, 100_rad, etc.)

PR #12, opened 2017-04-28
Timers and TimeWorldComponent

Liffes:
Hope this pull request works somehow, first time doing it.
Anyway, did standard timer that is updating constantly and pausable timer that doesn't tick when paused.
Also made TimeWorldComponent that has PausableTimer and updates it.


PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+//------------------------------------------------------------------------------
+void Poly::TimeSystem::TimeUpdatePhase(World * world)
+{
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
+
+	//Get new delta time
+	std::chrono::steady_clock::time_point frameTime = steady_clock::now();
+	duration<float> deltaTime = duration_cast<duration<float>>(frameTime - timeComponent->LastFrameTime);
+#ifdef DEBUG
```
MuniuDev: in future this should be surrounded with ifndef FINAL. As of right now, please remove it and add propper comment.

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+	for (auto& timer : timeComponent->Timers)
+	{
+		if (timeComponent->IsPaused && timer.second.IsPausable)
+			continue;
+		timer.second.Time += deltaTime;
+	}
+}
+
+//------------------------------------------------------------------------------
+void Poly::TimeSystem::RegisterTimer(World * world, const size_t id, const bool isPausable)
```
MuniuDev: Using const when passing argument by value is unneccessary and a bad practice. Please remove it. This applies to other methods as well.

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
+	//if timer with that id exists, do nothing (for now)
```
MuniuDev: When the same id is passed twice there should be an assertion. remove `if (timeComponent->Timers.count(id) == 0)` and replace it with `ASSERTE(timeComponent->Timers.count(id) == 0, "Same timer id provided twice!")`

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+//------------------------------------------------------------------------------
+float Poly::TimeSystem::GetDeltaTime(World * world)
+{
+	return world->GetTimeWorldComponent().GetDeltaTime();
+}
+
+//------------------------------------------------------------------------------
+float Poly::TimeSystem::GetTimerTime(World * world, const size_t id)
```
MuniuDev: `GetTimerElapsedTime` ?

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
+	//if timer with that id exists, do nothing (for now)
+	if (timeComponent->Timers.count(id) == 0)
+	{
+		timeComponent->Timers[id] = Timer(isPausable);
+	}
+}
+
+//------------------------------------------------------------------------------
+float Poly::TimeSystem::GetDeltaTime(World * world)
```
MuniuDev: BTW Delta time should also be part of Timer. Consider example when one timer has multiplier set to 0.5, it's delta time is different. Let's rename it to `GetTimerDeltaTime` and fix the implementation.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
...
+		float GetGameplayTime() const { return GameplayTime.GetTime(); };
+
+		void SetPaused(bool isPaused) { IsPaused = isPaused; };
+
+	private:
+		const milliseconds MIN_TIME_STEP_MS = milliseconds(60);
+
+		steady_clock::time_point LastFrameTime;
+
+		Timer SystemTime;
```
MuniuDev: Since we have `std::unordered_map<size_t, Timer> Timers`, `SystemTime` and `GameplayTime` should be also place inside it. They should be accessible by `enum class eEngineTimer { SYSTEM, GAMEPLAY, _COUNT }`. we can create helper methods in system namespace to access timers by `eEngineTimer` instead of `size_t`.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
...
+		Timer GameplayTime;
+
+		std::unordered_map<size_t, Timer> Timers;
+
+		float DeltaTime;
```
MuniuDev: please move delta time to timers.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
...
+	class ENGINE_DLLEXPORT TimeWorldComponent : public ComponentBase
+	{
+		friend void TimeSystem::TimeUpdatePhase(World* world);
+		friend void TimeSystem::RegisterTimer(World* world, const size_t id, const bool isPausable);
+		friend float TimeSystem::GetTimerTime(World* world, const size_t id);
+	public:	
+		TimeWorldComponent();
+		~TimeWorldComponent() {};
+
+		float GetDeltaTime() const { return DeltaTime; };
```
MuniuDev: These methods should be removed.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
...
+		float GetSystemTime() const { return SystemTime.GetTime(); };
+		float GetGameplayTime() const { return GameplayTime.GetTime(); };
+
+		void SetPaused(bool isPaused) { IsPaused = isPaused; };
```
MuniuDev: Let's add getter for checking if it's paused. Remember about const correctness.

PolyEngine/Engine/Src/Timer.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <chrono>
+#include <Core.hpp>
+#include "TimeSystem.hpp"
+
+using std::chrono::duration;
```
MuniuDev: never place `using` in header files.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include <chrono>
+#include <unordered_map>
+#include "TimeSystem.hpp"
+#include "Timer.hpp"
+#include "ComponentBase.hpp"
+
+using std::chrono::steady_clock;
```
MuniuDev: Please never place using in header files.

PolyEngine/Engine/TimeSystem.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <chrono>
+#include <Core.hpp>
+
+
+namespace Poly
+{
+	class ENGINE_DLLEXPORT TimeSystem : public BaseObject<>
```
MuniuDev: This file should be completely removed.

PolyEngine/Engine/Src/TimeSystem.cpp
VukoDrakkeinen: I'd argue that `const by default` is actually a *good* practice. Makes it easier to reason about the code. Now returning const values on the other hand...

PolyEngine/Engine/Src/TimeWorldComponent.hpp
VukoDrakkeinen: (unless it is scoped)

PolyEngine/Engine/TimeSystem.hpp
```diff
@@ -0,0 +1,30 @@
...
+	public:
+
+		TimeSystem();
+		~TimeSystem() {};
+
+		float Update();
+		float GetDeltaTime() const { return DeltaTime.count(); };
+		float GetPausableTimer() const { return PausableTimer.count(); };
+		float GetTimer() const { return Timer.count(); }
+
```
VukoDrakkeinen: Don't know how much do we care, but game time should actually be of `double` type. Float's precision is simply too low. After the game has been running for a day the precision drops to 7.8 milliseconds, and after a week to 62.5 milliseconds! (admittedly that's a pretty rare case) And then you've got stuff like occasional catastrophic cancellations when calculating delta time.

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+using std::chrono::duration;
+
+//------------------------------------------------------------------------------
+void Poly::TimeSystem::TimeUpdatePhase(World * world)
+{
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
+
+	//Get new delta time
+	std::chrono::steady_clock::time_point frameTime = steady_clock::now();
+	duration<float> deltaTime = duration_cast<duration<float>>(frameTime - timeComponent->LastFrameTime);
```
VukoDrakkeinen: Don't know how much do we care, but game time should actually be of `double` type. Float's precision is simply too low. After the game has been running for a day the precision drops to 7.8 milliseconds, and after a week to 62.5 milliseconds! (admittedly that's a pretty rare case) And then you've got stuff like occasional catastrophic cancellations when calculating delta time.

PolyEngine/Engine/Src/TimeSystem.cpp
VukoDrakkeinen: Should I add the define to our CMake script?

PolyEngine/WindowsGLStandalone/Src/Main.cpp
```diff
@@ -5,6 +5,7 @@
 #include <Engine.hpp>
 #include <OpenGLRenderingContext.hpp>
 #include <TestGame.hpp>
+#include <string>
```
VukoDrakkeinen: Why not use our string implementation?

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <chrono>
+
+using std::chrono::steady_clock;
+using std::chrono::duration_cast;
+using std::chrono::duration;
+
+//------------------------------------------------------------------------------
+void Poly::TimeSystem::TimeUpdatePhase(World * world)
+{
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
```
VukoDrakkeinen: Why convert reference to a pointer?

PolyEngine/Engine/Src/TimeSystem.cpp
```diff
@@ -0,0 +1,60 @@
...
+		timer.second.Time += deltaTime;
+	}
+}
+
+//------------------------------------------------------------------------------
+void Poly::TimeSystem::RegisterTimer(World * world, const size_t id, const bool isPausable)
+{
+	TimeWorldComponent* timeComponent = &world->GetTimeWorldComponent();
+	//if timer with that id exists, do nothing (for now)
+	if (timeComponent->Timers.count(id) == 0)
```
VukoDrakkeinen: Doesn't matter much, but it would be probably slightly faster to do `unordered_map::insert(id, timer)` and then check the result (if the key already exists in map, you get a bool of `false` value)

PolyEngine/Engine/Src/TimeWorldComponent.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+	private:
+		const milliseconds MIN_TIME_STEP_MS = milliseconds(60);
+
+		steady_clock::time_point LastFrameTime;
+
+		Timer SystemTime;
+		Timer GameplayTime;
+
+		std::unordered_map<size_t, Timer> Timers;
```
VukoDrakkeinen: Man, this `size_t` type is confusing... Wouldn't `UniqueId` be better here?

PolyEngine/Engine/Src/TimeSystem.cpp
MuniuDev: I agree, we should keep all calculations in timers in double. Actually it would be nice to have simple wrapper class for Time and TimePoint.

PolyEngine/WindowsGLStandalone/Src/Main.cpp
MuniuDev: Didn't notice it. Yes, usage of std::String is forbidden.

PolyEngine/Engine/Src/TimeSystem.cpp
MuniuDev: Just keep it as reference.

PolyEngine/Engine/Src/TimeWorldComponent.hpp
MuniuDev: UniqueID is right now only reserved to Entities. It will, in future, probably expand to resources and other serializable objects (to help pointer serialization) but i'm pretty sure it's not necessary here.

PolyEngine/Engine/TimeSystem.hpp
Liffes: whoops, that file must have been revived during rebase

PolyEngine/WindowsGLStandalone/Src/Main.cpp
Liffes: Just used it for a temporary visualisation of delta time, so i didn't put much tought into it, but no problem i guess

PolyEngine/Engine/Src/TimeWorldComponent.hpp
VukoDrakkeinen: Not necessary, sure, but imho even a type alias would tremendously help with readability.

PolyEngine/WindowsGLStandalone/Src/Main.cpp
```diff
@@ -107,11 +107,12 @@ int WINAPI WinMain(HINSTANCE hInstance,
 		{
 			// Run game code here
 			Engine.Update();
-
 			// Temporary visualisation of delta time in window title
-			float dTime = Engine.GetWorld().GetTimeWorldComponent().GetDeltaTime();
-			std::string windowText = "DeltaTime = " + std::to_string(dTime);
-			SetWindowText(hWnd, windowText.c_str());
+			std::stringstream ss;
```
VukoDrakkeinen: Uh, oh. It seems we could use a `toString()` function :P And a complete String implementation+API, right now it's pretty bare-bones.

MuniuDev: Yes, this will be one of the objectives of the next sprint.

PR #13, opened 2017-04-28
Linux port

VukoDrakkeinen:
Support for both GCC and Clang (MSVC coming soon™)

Now with 431% more single-line fixes!
 Fixed lots of minor issues as pointed out by GCC, Clang,
 Clang sanitizers and clang-tidy.

Includes CMake build scripts
Requires X11 with GLX (Wayland not supported)

Removes some unused includes

I had quite a few doubts about some things. You'll probably notice
them easily. If something looks weird, then that's probably it :D
Also we should probably split platform-dependent code into separate
files that would be then ifdef-included. The way it is now, it's
easier to merge, so I will probably make another PR.

P.S. Have mercy. I tried to limit the changes to minimum (the diff is a bit busy due to whitespace stripping though).


PolyEngine/Core/Src/String.hpp
```diff
@@ -6,21 +6,13 @@
...
-
-	template<typename T>
 	class StringBase : BaseObject<> {
 	public:
 		StringBase() { Data.Resize(1); Data[0] = 0; }
 
 		//------------------------------------------------------------------------------
 		StringBase(const T* data) {
-			size_t length = StrLen(data);
+			size_t length = std::strlen(data);
```
MuniuDev: This won't work. std::strlen has onlu one overload: `std::size_t strlen( const char* str );` which prevents it from being used in templated structure.

PolyEngine/Core/Src/UniqueID.cpp
```diff
@@ -1,7 +1,5 @@
 #include "CorePCH.hpp"
 
-#include "UniqueID.hpp"
```
MuniuDev: Please leave this header.

PolyEngine/Core/Src/Vector.cpp
```diff
@@ -1,9 +1,5 @@
 #include "CorePCH.hpp"
 
-#include "Vector.hpp"
```
MuniuDev: Please leave Vector.hpp

PolyEngine/Engine/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+find_package(OpenGL REQUIRED)
+if(WIN32)
+	find_package(GLEW REQUIRED PATHS ThirdParty/glew/lib/Release/ PATH_SUFFIXES Win32 x64) #FIXME(vuko): does this even work? how to make it search default paths first? (according to docs it should by default, but it fails on my system as opposed to invocation without given paths)
```
MuniuDev: Usually you have to set certain variable. I have and example here (https://github.com/MuniuDev/SpaceRush/blob/master/CMakeLists.txt) of working multi OS cmake with glew and other stuff.

PolyEngine/Engine/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+	Src/RenderingSystem.cpp
+	Src/ShaderProgram.cpp
+	Src/TransformComponent.cpp
+	Src/ViewportWorldComponent.cpp
+)
+set(POLYENGINE_INCLUDE
+	Src
+)
+set(POLYENGINE_H_FOR_IDE
+	Src/CameraComponent.hpp
```
MuniuDev: can't we just search the whole folder?

PolyEngine/Engine/Engine.vcxproj
```diff
@@ -167,9 +167,7 @@
   <ItemGroup>
     <ClCompile Include="Src\CameraComponent.cpp" />
     <ClCompile Include="Src\CameraSystem.cpp" />
-    <ClCompile Include="Src\ComponentBase.cpp" />
     <ClCompile Include="Src\Engine.cpp" />
-    <ClCompile Include="Src\EnginePCH.cpp">
```
MuniuDev: This breaks windows build.

PolyEngine/Engine/Engine.vcxproj
```diff
@@ -167,9 +167,7 @@
   <ItemGroup>
     <ClCompile Include="Src\CameraComponent.cpp" />
     <ClCompile Include="Src\CameraSystem.cpp" />
-    <ClCompile Include="Src\ComponentBase.cpp" />
```
MuniuDev: Why this was removed?

PolyEngine/Engine/Engine.vcxproj.filters
```diff
@@ -47,9 +47,6 @@
     <ClCompile Include="Src\World.cpp">
       <Filter>Source Files\ECS</Filter>
     </ClCompile>
-    <ClCompile Include="Src\ComponentBase.cpp">
-      <Filter>Source Files\ECS</Filter>
-    </ClCompile>
```
MuniuDev: See previous comments.

PolyEngine/Engine/Engine.vcxproj.filters
```diff
@@ -74,9 +71,6 @@
     <ClCompile Include="Src\InputWorldComponent.cpp">
       <Filter>Source Files\Input</Filter>
     </ClCompile>
-    <ClCompile Include="Src\EnginePCH.cpp">
-      <Filter>Source Files</Filter>
-    </ClCompile>
```
MuniuDev: See previous comments.

PolyEngine/Engine/Src/CameraComponent.cpp
```diff
@@ -1,7 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include "CameraComponent.hpp"
```
MuniuDev: Leave it.

PolyEngine/Engine/Src/CameraSystem.cpp
```diff
@@ -1,9 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include <set>
-
-#include "CameraSystem.hpp"
```
MuniuDev: Leave it.

PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -22,18 +17,18 @@ Poly::Engine::~Engine()
...
-bool Engine::Init(const IRenderingContextParams* context) 
+bool Engine::Init(const IRenderingContextParams* context)
 {
 	// Engine Components
-	REGISTER_COMPONENT(this, TransformComponent, eEngineComponents::TRANSFORM);
-	REGISTER_COMPONENT(this, CameraComponent, eEngineComponents::BASE_CAMERA);
-	REGISTER_COMPONENT(this, MeshRenderingComponent, eEngineComponents::MESH_RENDERING);
+	this->RegisterComponent<TransformComponent>((size_t) eEngineComponents::TRANSFORM);
+	this->RegisterComponent<CameraComponent>((size_t) eEngineComponents::BASE_CAMERA);
+	this->RegisterComponent<MeshRenderingComponent>((size_t) eEngineComponents::MESH_RENDERING);
```
MuniuDev: remove `this->`

PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -58,7 +53,7 @@ void Engine::RegisterUpdatePhase(const PhaseUpdateFunction& phaseFunction, eUpda
 }
 
 //------------------------------------------------------------------------------
-void Engine::Update(float dt)
+void Engine::Update(float /*dt*/)
```
MuniuDev: actually dt should be removed alltogether

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -11,9 +11,8 @@
 
 #include "InputSystem.hpp"
 
-namespace Poly 
+namespace Poly
 {
-	class World;
```
MuniuDev: How this file compile if `class World` was removed? This is undefined symbol now here `World& GetWorld() { return *BaseWorld; }` and here `World* BaseWorld;`

Don't know if this break linux build but it definetly breaks Windows one.

PolyEngine/Engine/Src/EnginePCH.hpp
```diff
@@ -7,8 +7,8 @@
 #include "Engine.hpp"
 
 // ECS
-#include "ComponentBase.hpp"
 #include "Entity.hpp"
+#include "ComponentBase.hpp"
```
MuniuDev: Why?

PolyEngine/Engine/Src/Entity.cpp
```diff
@@ -1,7 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include "Entity.hpp"
```
MuniuDev: See other comments.

PolyEngine/Engine/Src/Entity.hpp
```diff
@@ -48,13 +47,7 @@ namespace Poly
...
 		//////////////////////////////
 		template<class T>
-		T* GetComponent()
-		{
-			if (HasComponent(GET_COMPONENT_ID(EntityWorld->GetEngine(), T)))
-				return static_cast<T*>(Components[GET_COMPONENT_ID(EntityWorld->GetEngine(), T)]);
-			else
-				return nullptr;
-		}
+		T* GetComponent(); //defined in World.hpp due to circular inclusion problem
```
MuniuDev: This is bad resolution of this problem. Also i don't see the circular inclussion. please elaborate.

PolyEngine/Engine/Src/InputSystem.cpp
```diff
@@ -1,7 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include "InputSystem.hpp"
```
MuniuDev: Leave It

PolyEngine/Engine/Src/InputWorldComponent.cpp
```diff
@@ -1,7 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include "InputWorldComponent.hpp"
```
MuniuDev: Leave It

PolyEngine/Engine/Src/MeshRenderingComponent.cpp
```diff
@@ -1,8 +1,5 @@
 #include "EnginePCH.hpp"
 
-#include "CameraComponent.hpp"
-#include "MeshRenderingComponent.hpp"
```
MuniuDev: Leave It

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
```diff
@@ -22,30 +28,30 @@ bool OpenGLRenderingContext::Init(const IRenderingContextParams * context)
 {
 	const OpenGLRenderingContextParams* openGLContext = static_cast<const OpenGLRenderingContextParams*>(context);
 	hWnd = openGLContext->HWnd; // Set the HWND for our window  
-	hDC = GetDC(hWnd); // Get the device context for our window  
+	hDC = GetDC(hWnd); // Get the device context for our window  ScreenDim.Width = openGLContext->Rect.right - openGLContext->Rect.left;
```
MuniuDev: What is this?

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
```diff
@@ -127,3 +133,94 @@ const ScreenSize& Poly::OpenGLRenderingContext::GetScreenSize() const
...
+
+#elif defined(__linux__)
+
+//------------------------------------------------------------------------------
+bool OpenGLRenderingContext::Init(const IRenderingContextParams* ictxParams) {
+	auto ctxParams = static_cast<const OpenGLRenderingContextParams*>(ictxParams);
+	this->display = ctxParams->display;
+	this->window = ctxParams->window;
+
+	//what the fuck, GLEW
```
MuniuDev: remove this comment.

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
```diff
@@ -127,3 +133,94 @@ const ScreenSize& Poly::OpenGLRenderingContext::GetScreenSize() const
...
+	}
+
+	if (glXIsDirect(this->display, this->context)) {
+		Poly::gConsole.LogDebug("Direct GLX rendering context obtained");
+	} else {
+		Poly::gConsole.LogDebug("Indirect GLX rendering context obtained");
+	}
+	glXMakeCurrent(this->display, this->window, this->context);
+
+	Poly::gConsole.LogInfo("OpenGL context set up succesfully");
```
MuniuDev: we are using namespace Poly already.

PolyEngine/Engine/Src/ShaderProgram.cpp
```diff
@@ -123,5 +123,5 @@ void ShaderProgram::SetUniform(const String& name, const Vector& val)
 
 void ShaderProgram::SetUniform(const String& name, const Matrix& val)
 {
-	glUniformMatrix4fv(m_uniforms[name], 1, GL_FALSE, val.GetTransposed().Data);
+	glUniformMatrix4fv(m_uniforms[name], 1, GL_FALSE, val.GetTransposed().Data.data());
```
MuniuDev: please create `float* GetDataPtr()` method in all math types changed by this PR to make this more convinient

PolyEngine/Engine/Src/World.hpp
```diff
@@ -76,7 +76,7 @@ namespace Poly {
 		template<typename T>
 		T* GetComponent(const UniqueID& entityId)
 		{
-			HEAVY_ASSERTE(entityId < MAX_COMPONENTS_COUNT, "Invalid component ID");
+			HEAVY_ASSERTE(!!entityId, "Invalid component ID");
```
MuniuDev: good change. Please change also assertion message.

PolyEngine/UnitTests/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+	)
+	add_library(Catch INTERFACE)
+	add_dependencies(Catch Catch-External)
+	target_include_directories(Catch INTERFACE ${CMAKE_BINARY_DIR}/external/Catch/include)
+endif(NOT TARGET Catch)
+
+add_executable(polytests ${POLYTESTS_SRCS})
+target_link_libraries(polytests polycore polyengine Catch)
+#TODO(vuko): you have to manually add a case here every time you add a new test to Catch... Maybe grep the files looking for TEST_CASE then generate?
+add_test(NAME "Pool-allocator"                                COMMAND polytests "Pool allocator")
```
MuniuDev: Does this mean that invoking build also runs unit tests? What is the benefint of that?

PolyEngine/UnitTests/Src/EnumUtilsTests.cpp
```diff
@@ -5,15 +5,17 @@
 
 using namespace Poly;
 
-enum class eTestEnum
+enum class eTestEnum : uint8_t
```
MuniuDev: why? There is completely no need for that.

PolyEngine/Engine/Src/KeyBindings.hpp
```diff
@@ -1,11 +1,12 @@
 #pragma once
 
+#if defined(_WIN32)
 #include <Windows.h>
 
 namespace Poly
 {
-	enum class eKey
-	{	
+	enum class eKey : uint8_t
```
MuniuDev: Why do we want to have base type of uint8_t? This introduces great risk of overflow.

PolyEngine/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang") #TODO(vuko): cleanup, build types (debug, etc)
+	add_compile_options(-pipe)
+	add_compile_options(-Wall)
+	add_compile_options(-Wextra)
+	#add_compile_options(-stdlib=libc++)
+	#add_compile_options(-lc++abi)
+	if(SIMD)
+		add_compile_options(-msse3)
+	endif(SIMD)
+elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
VukoDrakkeinen: Yes, as stated I want to add the support in my next PR (get the changes in first, then iterate on that to improve)

PolyEngine/Core/Core.vcxproj
```diff
@@ -130,9 +130,7 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClCompile Include="Src\BaseObject.cpp" />
```
VukoDrakkeinen: Because it was empty (baring the single include) and useless. The implementation is wholly contained in the header file (maybe we should move template implementations to .tpp files?)

PolyEngine/Core/Core.vcxproj
```diff
@@ -130,9 +130,7 @@
...
     <ClCompile Include="Src\Color.cpp" />
-    <ClCompile Include="Src\CorePCH.cpp">
```
VukoDrakkeinen: This is weird, precompiled headers should not need .cpp files... Could you give me some resources to read on MSVC support for precompiled headers?

PolyEngine/Core/Core.vcxproj.filters
```diff
@@ -37,18 +37,12 @@
     <ClCompile Include="Src\Color.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="Src\BaseObject.cpp">
-      <Filter>Source Files\Memory</Filter>
-    </ClCompile>
```
VukoDrakkeinen: Why? The file's not there anymore.

PolyEngine/Core/Core.vcxproj.filters
```diff
@@ -37,18 +37,12 @@
...
     <ClCompile Include="Src\Logger.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
     <ClCompile Include="Src\UniqueID.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="Src\CorePCH.cpp">
-      <Filter>Source Files</Filter>
-    </ClCompile>
```
VukoDrakkeinen: As above

PolyEngine/Core/Src/AARect.cpp
```diff
@@ -1,5 +1,4 @@
 #include "CorePCH.hpp"
-#include "AARect.hpp"
```
VukoDrakkeinen: Slightly weird, but alright

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -4,8 +4,87 @@
...
-	//FIXME _aligned_malloc is Visual Studio only
+#if defined(_WIN32)
 	inline void* DefaultAlloc(size_t size) { return _aligned_malloc(size, 16); }
 	inline void* DefaultRealloc(void* ptr, size_t size) { return _aligned_realloc(ptr, size, 16); }
 	inline void DefaultFree(void* ptr) { _aligned_free(ptr); }
-}
\ No newline at end of file
+#elif defined(__GNUC__) || defined(__clang__)
+	namespace Impl {
+		constexpr size_t alignment = 16;
```
VukoDrakkeinen: Right, my mistake

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -4,8 +4,87 @@
...
+		return fresh;
+	}
+
+	inline void* DefaultRealloc(void* ptr, size_t size) {
+		using namespace Impl;
+		//POSIX provides no aligned_realloc, so always do a fresh allocation, then copy :(
+		void* fresh = DefaultAlloc(size);
+
+		if (ptr) {
+			memcpy(fresh, ptr, malloc_usable_size(ptr));
```
VukoDrakkeinen: Yes, as fas as I know, aligned_malloc() has the same semantics as malloc()

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -4,8 +4,87 @@
...
+			free(ptr);
```
VukoDrakkeinen: Noted

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -4,8 +4,87 @@
...
+		}
+
+		return fresh;
+	}
+
+	inline void DefaultFree(void* ptr) {
+		free(ptr);
```
VukoDrakkeinen: No such function, aligned_alloc'd memory works with free()

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -4,8 +4,87 @@
...
+	}
+
+#if 0
```
VukoDrakkeinen: Technically malloc_usable_size() should only be used for debugging, which is why I included another implementation that uses manual bookkeeping instead (and uses some additional memory for that). I was hesitant which implementation to choose.

PolyEngine/Core/Src/Color.cpp
```diff
@@ -1,7 +1,5 @@
 #include "CorePCH.hpp"
 
-#include "Color.hpp"
```
VukoDrakkeinen: Is this like the AARect one?

PolyEngine/Core/Src/Defines.hpp
```diff
@@ -52,6 +56,24 @@
 	#else
 		#define GAME_DLLEXPORT __declspec(dllimport)
 	#endif
+#elif defined(__GNUC__) || defined(__clang__)
+	#if defined(_DLL) && defined(_CORE)
+		#define CORE_DLLEXPORT __attribute__ ((visibility ("default")))
+	#else
```
VukoDrakkeinen: Made to parallel the Windows one. Will flatten.

PolyEngine/Core/Src/EnumUtils.hpp
```diff
@@ -64,7 +65,8 @@ namespace Poly {
 			friend class EnumArray<T, E, SIZE>;
 		};
 
-		EnumArray() = default;
+		//EnumArray() = default;
+		EnumArray() : Data{} {}
```
VukoDrakkeinen: Uninitialized memory read

PolyEngine/Core/Src/EnumUtils.hpp
```diff
@@ -126,18 +128,22 @@ namespace Poly {
...
 
-
-#define REGISTER_ENUM_NAMES(type, ...) template<> struct Poly::Impl::EnumInfo<type> { \
-	static Poly::Impl::EnumInfo<type>& Get() {static Poly::Impl::EnumInfo<type> instance; return instance; } \
-	const Poly::EnumArray<const char*, type> Names = {__VA_ARGS__}; }
\ No newline at end of file
+//NOTE(vuko): apparently defining specializations in a namespace from global/other namespace is illegal C++ and GCC complains
+//Unfortunately being compliant causes problems when using the macro in a namespace. If you run into problems, close the namespace, use the macro, then open it again
+#define REGISTER_ENUM_NAMES(type, ...)                                                    \
+	namespace Poly { namespace Impl {                                                     \
```
VukoDrakkeinen: Yes, I spent a lot of time on this trying to get it work without workarounds... Right now I close the namespace, use the macro, then open it again. It works, but two macros seems neater.

PolyEngine/Core/Src/Logger.cpp
```diff
@@ -1,7 +1,5 @@
 #include "CorePCH.hpp"
 
-#include "Logger.hpp"
```
VukoDrakkeinen: Just making sure, same as AARect?

PolyEngine/Core/Src/Logger.hpp
```diff
@@ -8,9 +8,13 @@ namespace Poly {
 	/**
 	*  Enum describing possible levels of logging
 	*/
-	enum class eLogLevel { LVL_DEBUG, LVL_INFO, LVL_WARNING, LVL_ERROR, _COUNT };
-	REGISTER_ENUM_NAMES(eLogLevel, "DEBUG", "INFO", "WARNING", "ERROR");
+	enum class eLogLevel : uint8_t { LVL_DEBUG, LVL_INFO, LVL_WARNING, LVL_ERROR, _COUNT };
+} //namespace Poly
 
+//WORKAROUND(vuko): namespaces specializations problem, see EnumUtils.hpp
+REGISTER_ENUM_NAMES(Poly::eLogLevel, "DEBUG", "INFO", "WARNING", "ERROR");
```
VukoDrakkeinen: I know :(

PolyEngine/Core/Src/Matrix.cpp
```diff
@@ -1,18 +1,13 @@
 #include "CorePCH.hpp"
 
-#include "Matrix.hpp"
```
VukoDrakkeinen: Assuming it's the AARect issue again

PolyEngine/Core/Src/Matrix.hpp
```diff
@@ -141,10 +141,13 @@ namespace Poly {
 	  // This structure allows to access vector elements by index or name.
 	  union {
 	#if !DISABLE_SIMD
-		__m128 SimdRow[4];
+	#pragma GCC diagnostic push
```
VukoDrakkeinen: Unknown pragmas are ignored, but I forgot they can emit warnings. Good catch, I will add an ifdef-guard

PolyEngine/Core/Src/String.hpp
VukoDrakkeinen: Ooh, so that's what it was! It could've used a comment. Why do we need anything other than `char` though?

PolyEngine/Core/Src/UniqueID.cpp
VukoDrakkeinen: AARect? Alright.

PolyEngine/Core/Src/Vector.cpp
VukoDrakkeinen: Still making sure, AARect?

PolyEngine/Engine/CMakeLists.txt
VukoDrakkeinen: Oh my, that is one clean CMake script; I could learn from that. Well, what I wanted to do is to search the system path first, then fall back on the included one. And since the CMake docs confuse the hell out of me, this is the result. I'm still not sure about the given example, though, please advise.

PolyEngine/Engine/CMakeLists.txt
VukoDrakkeinen: GLOB? No, unless you want to `touch` the CMakeLists.txt every time you add a header (so CMake can regenerate the list)

PolyEngine/Engine/Engine.vcxproj
VukoDrakkeinen: Will revert. Still think the MSVC PCH implementation is weird as hell.

PolyEngine/Engine/Src/Engine.cpp
VukoDrakkeinen: This makes me sad, but alas I will obey!

PolyEngine/Engine/Src/Engine.cpp
VukoDrakkeinen: Alright.

PolyEngine/Engine/Src/Engine.hpp
VukoDrakkeinen: Oooh, good catch. It works, because it's forward-declared in InputSystem.hpp, which is included here. It's weird that I removed it though, must have been a mistake.

PolyEngine/Engine/Src/EnginePCH.hpp
VukoDrakkeinen: Does this not belong here? What are the rules for PCH inclusion, then?

PolyEngine/Engine/Src/Entity.hpp
VukoDrakkeinen: (explained on Slack)

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
VukoDrakkeinen: Oh, so that's where the line went! Had to manually re-add it (and missed that it was still there, just in the comment). Will fix.

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
VukoDrakkeinen: :(

PolyEngine/Engine/Src/OpenGLRenderingContext.cpp
VukoDrakkeinen: Right. Copy-paste programming™ strikes again...

PolyEngine/Engine/Src/World.hpp
VukoDrakkeinen: I was actually wondering about that. What should I change it to?

PolyEngine/UnitTests/CMakeLists.txt
VukoDrakkeinen: No this just adds our tests to CTest, so one can run `make tests` and get a sane output. You can still manually run the resulting executable though, so it's mostly a convenience thing.

PolyEngine/Engine/Src/KeyBindings.hpp
VukoDrakkeinen: I actually checked and all our values are within range. As far as I know, there are no keycodes higher than 255 in the wild (although both WinAPI and X11 use an int for storing them, so I should probably change it).

PolyEngine/Core/Core.vcxproj
MuniuDev: Well in MSVC they do. http://www.asawicki.info/productions/artykuly/PrecompiledHeaders.php5

PolyEngine/Core/Core.vcxproj
MuniuDev: Ok, i've forgot this is template only class.

PolyEngine/Core/Src/Allocator.hpp
MuniuDev: Well, keep in mind that the address returned from aligned malloc is not neccesairly the address of the block that is allocated by it.

PolyEngine/Core/Src/Allocator.hpp
MuniuDev: I'm not sure this is safe, i.e. on MSVC this will crash.

PolyEngine/Core/Src/Allocator.hpp
MuniuDev: Then please provide a comment. I think it's best to keep the malloc_usable_size() for now, and roll out our own implementation of aligned malloc in the future.

PolyEngine/Core/Src/Color.cpp
MuniuDev: yup, this applies to many other files.

PolyEngine/Core/Src/EnumUtils.hpp
MuniuDev: The defult contructor won't initialize the array of non-pod objects?

PolyEngine/Core/Src/String.hpp
MuniuDev: Well, for theoretical future utf-16 support. Not really sure if this will happen though.

PolyEngine/Engine/CMakeLists.txt
MuniuDev: it will regenerate everything, i've used it sucessfully, read example attached in previous comment.

PolyEngine/Engine/Src/EnginePCH.hpp
MuniuDev: You just moved it two lines down ;p

PolyEngine/Engine/Src/World.hpp
MuniuDev: Invalie entity ID? :P

PolyEngine/Engine/Src/KeyBindings.hpp
MuniuDev: It's always better to cach an assertion with value 257 than have it overflow to 1 and go completely silent.

PolyEngine/Core/Src/Allocator.hpp
VukoDrakkeinen: Well, I wouldn't be able write the manual-bookkeeping version of the code without knowing that, would I :P

PolyEngine/Core/Src/Allocator.hpp
VukoDrakkeinen: From the docs: `returns either NULL, or a unique pointer value that can later be successfully passed to free(3)`

PolyEngine/Core/Src/Allocator.hpp
VukoDrakkeinen: How about leveraging this `http://en.cppreference.com/w/cpp/types/aligned_storage`?

PolyEngine/Core/Src/EnumUtils.hpp
VukoDrakkeinen: Yeah, it will default-initialize them, which (for pointers) is doing nothing (as per http://en.cppreference.com/w/cpp/language/default_initialization).

PolyEngine/Core/Src/String.hpp
VukoDrakkeinen: Why would we need to support a broken encoding?

PolyEngine/Engine/CMakeLists.txt
VukoDrakkeinen: Yes, make (or other build systems whose scripts were generated by CMake for that matter) will not rerun CMake unless you modify CMakeLists.txt (or make it seem like it was modified by `touch`ing it), which is even more inconvenient, as it introduces weird bugs and compilation problems. I would strongly advise against using GLOB, as it only caused me problems in the past.

PolyEngine/Engine/Src/EnginePCH.hpp
VukoDrakkeinen: Oh, because it was causing compilation problems (undefined symbols)

PolyEngine/UnitTests/Src/EnumUtilsTests.cpp
VukoDrakkeinen: I discussed this on Slack. It's a failed attempt to enforce the enum values to remain in the indexing range (for EnumArray). I don't think it's actually doable without code-gen, though :|
Right now it kind of works as a reminder, but I'm not opposed to removing it.

PolyEngine/Engine/Src/KeyBindings.hpp
VukoDrakkeinen: You're completely right, I don't know what I was thinking.

PolyEngine/Core/Src/Allocator.hpp
VukoDrakkeinen: Yes, obviously. Additionally I tested it extensively and it never failed to return the correct value.

PolyEngine/Engine/Src/World.hpp
VukoDrakkeinen: Oh... Silly me. I hadn't noticed at all!

PolyEngine/Core/Src/Allocator.hpp
MuniuDev: Ok, that explains all.

PolyEngine/Core/Src/Allocator.hpp
MuniuDev: We can, but this is subject for future PR.

PolyEngine/Core/Src/String.hpp
MuniuDev: we for sure need unicode, and probably utf-8 is best, but utf-16 is still very popular and may be neede to support files not created by our engine but by some outside editor.

PolyEngine/Engine/CMakeLists.txt
MuniuDev: Ok, this is easy change, probably best to talk it through on the meeting.

PolyEngine/Engine/Src/EnginePCH.hpp
MuniuDev: Then the problem is somwhere else. Entity.hpp problably includes something that is required by ComponentBase.

PolyEngine/Core/Src/String.hpp
VukoDrakkeinen: Not arguing against UTF-8, that's a no-brainer nowadays (and it nicely fits into chars!). Not really happy about allowing UTF-16, but alright. Out of curiosity, who still uses it?

PolyEngine/Engine/CMakeLists.txt
VukoDrakkeinen: I could always maintain the CMakeLists.txt by myself. I don't mind the occasional breakage and besides, we're still using MSBuild as a primary build system, right?

PolyEngine/Engine/Src/EnginePCH.hpp
VukoDrakkeinen: Well, yeah, but I assumed that we're moving the majority of our includes to the PCH file. This turned out to be wrong, so I'll be reverting those changes shortly.

PolyEngine/Core/Src/EnumUtils.hpp
```diff
@@ -99,8 +97,14 @@ namespace Poly {
 		const T* GetData() const { return Data; }
 
 		//------------------------------------------------------------------------------
-		T& operator[](E idx) { HEAVY_ASSERTE(static_cast<IndexType>(idx) < GetSize(), "Index out of bounds!"); return Data[static_cast<IndexType>(idx)]; }
-		const T& operator[](E idx) const { HEAVY_ASSERTE(static_cast<IndexType>(idx) < GetSize(), "Index out of bounds!"); return Data[static_cast<IndexType>(idx)]; }
+		T& operator[](E enumValue) {
+			return const_cast<T&>(const_cast<const EnumArray&>(*this)[enumValue]); //as advised by Effective C++
```
MuniuDev: From http://en.cppreference.com/w/cpp/language/const_cast : 

> const_cast makes it possible to form a reference or pointer to non-const type that is actually referring to a const object or a reference or pointer to non-volatile type that is actually referring to a volatile object. Modifying a const object through a non-const access path and referring to a volatile object through a non-volatile glvalue results in **undefined behavior**.

Please don't overuse const_cast and rever this change.

VukoDrakkeinen: This is actually fully safe, as supported by Effective C++, Third Edition, by Scott Meyers and _extensive_ online research 😃

PR #21, opened 2017-05-06
Platform-independent allocator

VukoDrakkeinen:


PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -6,6 +6,34 @@
 
 namespace Poly {
 
+	template<typename T>
```
MuniuDev: Why is this templated. This could easily return void* like malloc does.

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -6,6 +6,34 @@
...
+	T* Allocate(size_t count)
+	{
+		using AlignedT = std::aligned_storage<sizeof(T), 16>;
```
MuniuDev: here aligned storage should have <size, 16> as template args.

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -6,6 +6,34 @@
...
+		T* fresh = new(std::nothrow) AlignedT[count];
```
MuniuDev: this would call non array new.

PolyEngine/Core/Src/Allocator.hpp
```diff
@@ -6,6 +6,34 @@
...
+		return fresh;
+	}
+
+
+	template<typename T>
+	void Free(T* memory)
+	{
+		delete[] memory;
```
MuniuDev: a here should be normal delete.

PR #23, opened 2017-05-10
GetResourcesAbsolutePath() temporary implementation

R4yRayR4y:
It is just temporary implementation. This will be replaced with some JSON config.

Should default resources directory and config path be changed to something else?


PolyEngine/Engine/Src/ResourceManager.cpp
```diff
@@ -10,27 +10,21 @@ using namespace Poly;
 const String& Poly::GetResourcesAbsolutePath()
 {
 	static String PATH;
+	static String defaultPath("../Engine/Res/");
+	static String assetsPath("AssetsPath.txt");
```
MuniuDev: Please make them both static const.

PolyEngine/Engine/Src/ResourceManager.hpp
```diff
@@ -18,7 +18,7 @@ namespace Poly
 	{
 	public:
 		//------------------------------------------------------------------------------
-		static T* Load(const String& relativePath)
+		static T* Load(const String& relativePath, bool absolute=true)
```
MuniuDev: add spaces between operator =

PR #26, opened 2017-05-14
Path

DevRayy:


PolyEngine/Core/Src/Path.cpp
```diff
@@ -0,0 +1,63 @@
...
+	String rootTmp;
+	if (Root.GetLength())
+	{
+		rootTmp = Root + DELIMITER;
+	}
+	return rootTmp + dirsTmp + DELIMITER + Filename;
+}
+
+//------------------------------------------------------------------------------
+const String Path::DELIMITER("\\");
```
VukoDrakkeinen: Dude, no... Just... no...

PolyEngine/Core/Core.vcxproj.filters
```diff
@@ -55,6 +55,9 @@
     <ClCompile Include="Src\BaseObject.cpp">
       <Filter>Source Files\Memory</Filter>
     </ClCompile>
+    <ClCompile Include="Src\Path.cpp">
```
VukoDrakkeinen: If you could additionally add the files to `Core/CMakeFiles.txt` that would be great. Otherwise the build will break again.

PolyEngine/Core/Src/Path.cpp
```diff
@@ -0,0 +1,63 @@
+#include "CorePCH.hpp"
+
+#include "Path.hpp"
+
+using namespace Poly;
+
+//------------------------------------------------------------------------------
+Path* Path::Build()
+{
+	return new Path();
```
VukoDrakkeinen: Uh, why is this heap-allocated? o_O

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include "Defines.hpp"
+#include "Dynarray.hpp"
+#include "String.hpp"
+
+namespace Poly {
+
+	class CORE_DLLEXPORT Path
+	{
+	public:
+		static Path* Build();
```
VukoDrakkeinen: A common convention is for a `build()` method to return the final _built_ value, so this might be a little confusing.

PolyEngine/Core/Src/Path.cpp
```diff
@@ -0,0 +1,63 @@
...
+			dirsTmp = s;
+		}
+	}
+
+	String rootTmp;
+	if (Root.GetLength())
+	{
+		rootTmp = Root + DELIMITER;
+	}
+	return rootTmp + dirsTmp + DELIMITER + Filename;
```
VukoDrakkeinen: This should probably, first of all, cache the result, so consecutive calls to `Get()` do not recalculate the value, and secondly pre-allocate the final string. Right now every additional path element causes an allocation.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+		Path& RootAt(const String& root);
+		Path& Dir(const String& dir);
+		Path& File(const String& file);
+
+		bool IsAbsolute() const { return Root.GetLength(); }
```
VukoDrakkeinen: This will break on non-Windows systems. `C:/whatever` is valid _relative_ path.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,31 @@
...
+		bool IsDir() const { return !Filename.GetLength(); }
+
+		String Get();
+
```
VukoDrakkeinen: I'd love some additional convenience methods, like e.g. `::New(startingValue)`, `::SetExtension()`, `operator+`. Also getters for root, path without filename, file stem, file extension, etc.

PolyEngine/Core/Src/String.hpp
```diff
@@ -48,7 +48,7 @@ namespace Poly {
 		}
 
 		//------------------------------------------------------------------------------
-		bool operator==(const T* str) {
+		bool operator==(const T* str) const {
```
R4yRayR4y: I'm not sure about this, but without this operators beeing const I could not compare Path.Get() with other Strings. So I made them const just like they are for Matrix class.

PolyEngine/UnitTests/Src/PathTests.cpp
```diff
@@ -0,0 +1,19 @@
...
+	REQUIRE(p.Get() == "C:\\Dir1\\Dir2\\File.ext");
+	REQUIRE(p.IsAbsolute());
+	REQUIRE(!p.IsDir());
+
+	Path p2 = Path::Init()->Dir("Dir1").Dir("Dir2");
+	REQUIRE(p2.Get() == "Dir1\\Dir2\\");
+	REQUIRE(!p2.IsAbsolute());
+	REQUIRE(p2.IsDir());
+
+}
```
R4yRayR4y: This needs some work, I will do it later

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "String.hpp"
+
+namespace Poly {
+
+	class CORE_DLLEXPORT Path
+	{
+	public:
+		static Path* Init();
+		//TODO when String.split() will be ready
+		static Path* FromString(const String& source);
```
MuniuDev: Instead of these two static methods just provide propper constructors.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+		Path& RootAt(const String& root);
```
MuniuDev: This method is actually unnecessary and system dependant. Dir(..) can be made to handle "C:\" or "/", or we can provide non-static FromString method that let's us do that. i.e.
`Path p; p.FromString("C:\absolute\path\to\resource").Dir("dir1").File("resource.txt");`

MuniuDev: There should also be a way to convert relative path to absolute. This will make possible using relative paths to create it and then passing absolute path somwhere else.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+		Path& Dir(const String& dir);
+		Path& File(const String& file);
+
+		bool IsAbsolute() const { return Root.GetLength(); }
+		bool IsDir() const { return !Filename.GetLength(); }
+
+		const String& Get();
```
MuniuDev: This should be called ToString

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+		const String& GetRoot() { return Root; }
```
MuniuDev: Unnecessary.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+		const String& GetFilename() { return Filename; }
+		//TODO when String.split() will be ready
+		//const String& GetFileExtension();
```
MuniuDev: `Path GetDirectory()` would be nice to have.

PolyEngine/Core/Src/Path.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+		static const String DELIMITER;
```
MuniuDev: This should be private

PR #31, opened 2017-05-22
AARect Contains() and IsCollidingWith() functionality.

vadordev:
As per _[Code] Bounding Boxes Collisions_ task.


PolyEngine/Core/Src/AARect.cpp
```diff
@@ -13,4 +13,37 @@ namespace Poly {
...
+		const float interDown = std::min(r1MaxZ, r2MaxZ);
+
+		if ((interLeft < interRight) && (interTop < interBottom) && (interUp < interDown))
+		{
+			if (intersection)
+				*intersection = AARect(Vector(interLeft, interTop, interUp), Vector(interRight - interLeft, interBottom - interTop, interDown - interUp));
+			return true;
+		}
+
+		return false;
```
MuniuDev: This is really overcomplicated, this can be done in single return statement.
Example from the internet:
`bool DoBoxesIntersect(Box a, Box b) {
  return (abs(a.x - b.x) * 2 < (a.width + b.width)) &&
         (abs(a.y - b.y) * 2 < (a.height + b.height));
}`

If we really want the rect created by clipping the other rect the we should provide other method that does it. Still i'm pretty sure this would be extra couple lines long ;)

PR #32, opened 2017-05-23
Queue implementation based on Dynarray

Althiena:


PolyEngine/Core/Src/Queue.hpp
```diff
@@ -1,25 +1,226 @@
...
+			for (const T& obj : list)
+				PushBack(obj);
+		}
+
+		void Enlarge() { Reserve((Capacity + 1) * 2); }
+
+		size_t Head = 0;
+		size_t Tail = 0;
+		size_t Size = 0;
+		size_t Capacity = 1;
```
MuniuDev: Capacity should be 0 on start. No need to allocate memory if queue won't be used (for some reason). This is working better when someone is calling Reserve just after calling empty constructor.

PolyEngine/UnitTests/Src/QueueTests.cpp
```diff
@@ -9,47 +9,48 @@ TEST_CASE("Queue tests", "[Queue]")
 	Queue<int> q;
 	Queue<int> r(q);
 
-	REQUIRE(q.Empty() == true);
-	REQUIRE(r.Empty() == true);
+	REQUIRE(q.IsEmpty() == true);
```
MuniuDev: It would be very nice to have more elaborate unit tests that check for example various constructors and assign operators, resizes and reserves.

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -79,7 +79,7 @@ namespace Poly
 		bool operator!=(const Queue<T>& rhs) const { return !(*this == rhs); };
 
 		//------------------------------------------------------------------------------
-		bool IsEmpty() const { return GetSize() == 0; }
+		bool Empty() const { return GetSize() == 0; }
```
MuniuDev: Why the rename? In Dynarray the method is called `IsEmpty()` let's revert this change and keep it consistent.

PolyEngine/Core/Src/Queue.hpp
```diff
@@ -96,42 +96,42 @@ namespace Poly
 		//------------------------------------------------------------------------------
 		void PushBack(const T& obj)
 		{
+			if (Empty() || Head == GetNextIdx(Tail))
```
MuniuDev: You can simply check if `GetSize() + 1 > GetCapacity()` here and in the other places.

PolyEngine/Core/Src/Queue.hpp
Althiena: You've written "make it empty" in last change requests, and I misunderstood you. Thought you want me to rename this function, but you meant to make default ctor empty ;P

PR #38, opened 2017-05-31
TransformComponent child transform propagation fix, fixes #37

### LINKED ISSUE #37: Child Transform Components not moving relative to parent's one.
Child Transform Components not moving relative to parent's one if parent's transform was calculated before ( UpdateGlobalTransformationCache() was called on parent first, then on child).

To replicate use following:
```
tc2.SetParent(&tc1);
    tc1.SetLocalTranslation(v1);
    tc1.GetGlobalTransformationMatrix();
    REQUIRE(tc2.GetGlobalTranslation() == v1);
    REQUIRE(tc2.GetGlobalTransformationMatrix().Data[3] == v1.X);
```

R4yRayR4y:


PolyEngine/Engine/Src/TransformComponent.cpp
```diff
@@ -129,29 +129,40 @@ bool TransformComponent::UpdateLocalTransformationCache() const
 }
 
 //------------------------------------------------------------------------------
-bool TransformComponent::UpdateGlobalTransformationCache() const
+void TransformComponent::UpdateGlobalTransformationCache() const
 {
+	if (!GlobalDirty) return;
 	if (Parent == nullptr)
 	{
 		if (GlobalDirty) {
```
MuniuDev: This whole if block is unnecessary, drop it.

PolyEngine/Engine/Src/TransformComponent.cpp
```diff
@@ -129,29 +129,40 @@ bool TransformComponent::UpdateLocalTransformationCache() const
...
+			return;
 		}
 		else
-			return false;
+			return;
 	}
 	else
 	{
-		if (Parent->UpdateGlobalTransformationCache() || GlobalDirty)
+		if (GlobalDirty)
```
MuniuDev: Again, drop this.

PR #42, opened 2017-06-07
Destroying children

PiotrMoscicki:


PolyEngine/Engine/Src/World.cpp
```diff
@@ -49,7 +49,7 @@ void World::DestroyEntity(const UniqueID& entityId)
 	TransformComponent* transform = ent->GetComponent<TransformComponent>();
 	if (transform)
 	{
-		for (auto it : transform->GetChildren()) DestroyEntity(it->GetOwnerID());
```
MuniuDev: You can and should call `DestroyEntity` method here it's private, but you are calling it from within the class.

PolyEngine/Engine/Src/World.hpp
```diff
@@ -139,6 +139,7 @@ namespace Poly {
 
 		DeferredTaskQueue& GetDeferredTaskQueue() { return DeferredTasksQueue; }
 
+		void DestroyEntity(const UniqueID& entityId);
```
MuniuDev: Make it back a private method.

PolyEngine/Engine/Src/World.cpp
```diff
@@ -45,6 +45,14 @@ void World::DestroyEntity(const UniqueID& entityId)
 {
 	Entity* ent = IDToEntityMap[entityId];
 	HEAVY_ASSERTE(ent, "Invalid entity ID");
+
+	TransformComponent* transform = ent->GetComponent<TransformComponent>();
+	if (transform)
+	{
+		for (auto it : transform->GetChildren()) DeferredTaskSystem::DestroyEntityImmediate(this, it->GetOwnerID());
+		transform->GetChildren().Clear();
```
MuniuDev: No need to clear that collection. It will be empty by the time all children are removed.

PR #56, opened 2017-08-06
Add an ordered map based on a B-Tree

VukoDrakkeinen:
Here it is (finally). On my PC it's consistently up to 3x faster than the std::map.
I tried hard to make the API as easy to use and code as readable as possible, but I can't tell if I did a satisfactory job or not. I'm expecting lots of back'n'forth, so I made this a RFC.


PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		Optional<V> Insert(const K&  key, const V&  value) { return this->insert(          key ,           value ); }
+		Optional<V> Insert(const K&  key,       V&& value) { return this->insert(          key , std::move(value)); }
+		Optional<V> Insert(      K&& key, const V&  value) { return this->insert(std::move(key),           value ); }
+		Optional<V> Insert(      K&& key,       V&& value) { return this->insert(std::move(key), std::move(value)); }
+
+		Optional<V> Remove(const K&  key) { return this->remove(key); }
+		Optional<V> Remove(      K&& key) { return this->remove(std::move(key)); }
+
+		Optional<V&> Get(const K&  key) { return this->get(key); }
+		Optional<V&> Get(      K&  key) { return this->get(std::move(key)); }
```
MuniuDev: Provide const overload for these two methods.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+		V& operator[](const K& key) { return this->Get(key).value(); }
+		size_t GetSize() { return this->len; };
+		bool IsEmpty()   { return this->GetSize() == 0; };
```
MuniuDev: Change these two to constant methods.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		Optional<V> Insert(      K&& key, const V&  value) { return this->insert(std::move(key),           value ); }
+		Optional<V> Insert(      K&& key,       V&& value) { return this->insert(std::move(key), std::move(value)); }
+
+		Optional<V> Remove(const K&  key) { return this->remove(key); }
+		Optional<V> Remove(      K&& key) { return this->remove(std::move(key)); }
+
+		Optional<V&> Get(const K&  key) { return this->get(key); }
+		Optional<V&> Get(      K&  key) { return this->get(std::move(key)); }
+
+		V& operator[](const K& key) { return this->Get(key).value(); }
```
MuniuDev: Provide constant overload.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+			//reinitialize the root node
+			new_root_to_be->~LeafNode();
+			::new(new_root_to_be) LeafNode();
+
+			this->root = {new_root_to_be, 0};
+		}
+
+	public:
+		Entry Entry(const K& key) {
```
MuniuDev: As we speaked earlier. It would be much better to replace Entry with proper iterators and find method.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		}
+		Range<K, V, Bfactor> MaxRange() { return {BTree::first_leaf_edge(this->root.as_node_ref()), BTree::last_leaf_edge(this->root.as_node_ref())}; }
+		Range<K, V, Bfactor> GetRange(const K& lower, const K& upper) { ASSERTE(lower < upper, ""); return {search_tree(this->root.as_node_ref(), lower).handle, search_tree(this->root.as_node_ref(), upper).handle}; }
+		Iter begin() { return {BTree::first_leaf_edge(this->root.as_node_ref()), this->GetSize(), this->GetSize()}; }
+		Iter end()   { return {BTree:: last_leaf_edge(this->root.as_node_ref()), 0,               this->GetSize()}; }
+		class Keys   Keys()   { return {*this}; }
+		class Values Values() { return {*this}; }
+
+	public:
+		void swap(BTreeMap& other) { std::swap(this->root, other.root); std::swap(this->len, other.len); }
```
MuniuDev: Start with capital letter.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+				case SearchResult::FAILED: return Entry::Vacant(  search_result.handle, this->len, key);
+				default: UNREACHABLE();
+			}
+		}
+		Range<K, V, Bfactor> MaxRange() { return {BTree::first_leaf_edge(this->root.as_node_ref()), BTree::last_leaf_edge(this->root.as_node_ref())}; }
+		Range<K, V, Bfactor> GetRange(const K& lower, const K& upper) { ASSERTE(lower < upper, ""); return {search_tree(this->root.as_node_ref(), lower).handle, search_tree(this->root.as_node_ref(), upper).handle}; }
+		Iter begin() { return {BTree::first_leaf_edge(this->root.as_node_ref()), this->GetSize(), this->GetSize()}; }
+		Iter end()   { return {BTree:: last_leaf_edge(this->root.as_node_ref()), 0,               this->GetSize()}; }
+		class Keys   Keys()   { return {*this}; }
+		class Values Values() { return {*this}; }
```
MuniuDev: Provide const overloads for these six methods. `Keys()` and `Values()` should return propper iterator ranges. Ofcourse that implies implementation of `KeyIterator` and `ValueIterator`.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+	public:
+		void swap(BTreeMap& other) { std::swap(this->root, other.root); std::swap(this->len, other.len); }
+
+	private:
+		template<typename Key, typename Val>
+		Optional<V> insert(Key&& key, Val&& value) {
```
MuniuDev: All methods should start with capital letters. If you want to emphasise their privacy define them with `Impl` suffix.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		class Entry	{
+		public:
+			static Entry Vacant  (KVERef kv_ref, size_t& len, const K&  key) { return Entry(kv_ref, len, key); }
+			static Entry Vacant  (KVERef kv_ref, size_t& len,       K&& key) { return Entry(kv_ref, len, std::move(key)); }
+			static Entry Occupied(KVERef kv_ref, size_t& len)                { return Entry(kv_ref, len); }
+
+		public:
+			bool IsVacant() { return static_cast<bool>(key); }
+
+			V& OrInsert(const V&  value) { return this->or_insert(value); }
```
MuniuDev: Please don't use `this->` unless necessary.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+			V& OrInsert(      V&& value) { return this->or_insert(std::move(value)); }
+
+			V& VacantInsert(const V&  value) { return this->vacant_insert(value); }
+			V& VacantInsert(      V&& value) { return this->vacant_insert(std::move(value)); }
+
+			V OccupiedInsert(const V&  value) { return this->occupied_insert(value); }
+			V OccupiedInsert(      V&& value) { return this->occupied_insert(std::move(value)); }
+
+			V& OccupiedGet() { ASSERTE(!this->IsVacant(), "Cannot read value from a vacant map entry!"); return this->kv_ref.node_ref.node->values[this->kv_ref.idx]; }
```
MuniuDev: This entry api is very unsafe and hard to use. This is candidate for removal or making `private`.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+			size_t& map_len;
+		};
+
+		struct KV {
+			const K& key;
+			V& value;
+			const KV* operator->() const { return this; } //note(vuko): needed by the iterator's own operator ->, since it returns KV by value
+		};
+
+		class Iter : public std::iterator<std::bidirectional_iterator_tag, KV> {
```
MuniuDev: Rename to `Iterator`, provide `ConstIterator`

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+namespace Poly {
+	template<typename V>
+	class Optional {
+	public:
+		using ValueType = V;
+	public:
+		Optional() : initialized(false), dummy{} {}
+		Optional(const V&  value) : initialized(true), value_storage(value) {}
+		Optional(      V&& value) : initialized(true), value_storage(std::move(value)) {}
```
VukoDrakkeinen: I find aligned lines to be of tremendous help when it comes to code readability and how easy it is to spot a mistake. I try to keep a good balance as not to format code in a weird way, but sometimes there's a border case like this. Should I change this?
P.S. find the error:
`
Optional<V&> Get(const K&  key) { return this->get(key); }
Optional<V&> Get(      K&  key) { return this->get(std::move(key)); }
`
(I just fixed it :P)

VukoDrakkeinen: I find aligned lines to be of tremendous help when it comes to code readability and how easy it is to spot a mistake. I try to keep a good balance as not to format code in a weird way, but sometimes there's a border case like this. Should I change this?
P.S. find the error:
```
Optional<V&> Get(const K&  key) { return this->get(key); }
Optional<V&> Get(      K&  key) { return this->get(std::move(key)); }
```
(I just fixed it :P)

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+		~Optional() { if (this->HasValue()) { this->Value().~V(); } }
```
VukoDrakkeinen: Honestly, I don't see a point in using them. They're thin wrappers, that as far as I can see, don't bring anything to the table.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+	public:
```
VukoDrakkeinen: Sure it is! ;) I place them as separators and so I don't have to scroll far up to check for visibility of a given class member. I'm not strongly opposed to removing them, though.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+		Optional(Optional&& other) {
```
VukoDrakkeinen: Intentionally omitted. One is not interested in the Optional itself, but the value it holds. I don't see the use-case here.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+		bool HasValue() const { return this->initialized; }
+		operator bool() const { return this->HasValue(); }
+
+	public:
+		const V&  Value() const& { ASSERTE(this->HasValue(), "Attempting to get a value from an empty optional"); return this->value_storage; }
+		      V&  Value() &      { ASSERTE(this->HasValue(), "Attempting to get a value from an empty optional"); return this->value_storage; }
+		      V&& Value() &&     { ASSERTE(this->HasValue(), "Attempting to get a value from an empty optional"); return std::move(this->value_storage); }
+
+	public:
+		V TakeValue() {
```
VukoDrakkeinen: I'm opposed to this change. I feel that `MoveValue` doesn't clearly express what this method is about (taking the value out) and furthermore can be confused for move semantics. While they're extremely similar, they're not the same thing. Moving leaves the value in a valid (indeterminate) state, whereas taking does not.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+	private:
+		bool initialized;
+		union {
+			V value_storage;
+			char dummy;
+		};
+	};
+
+	template<typename V>
+	class Optional<V&> {
```
VukoDrakkeinen: It sure is! The specializations may look the same, but they're actually different underlying implementations with the same API.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+		V ValueOr(V&& fallback) const && { if (this->HasValue()) { return std::move(*this->Value()); } return fallback; }
+
+	private:
+		V* value_storage;
+	};
+
+	//todo(vuko): this is pretty much copy-pasted from the main implementation (with the copying constructor removed); can we somehow make an alias?
+	//always moves in the contents
+	template<typename V>
+	class Optional<V&&> {
```
VukoDrakkeinen: It's a convenience specialization. `Optional<T&&>` is effectively the same as `Optional<T>`, but it forbids passing in lvalue references. Another thing to note is that the specialization makes
Optionals play nicer with forwarding references. Consider this:
```
template<typename T>
void func(T&& t) {
	Optional<T>(std::forward<T>(t));
	//...do stuff with the optional
}
```
When passing a lvalue reference, no copying takes place, yet rvalue references are moved in as expected.

PolyEngine/Core/Src/Optional.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+	private:
+		bool initialized;
+		union {
+			V value_storage;
+			char dummy;
+		};
+	};
+
+	//todo(vuko): Optional-creating utility functions
```
VukoDrakkeinen: That ugly thing? :P I fail to see how it is useful, when one can simply use `{}`.
```
Optional<T> func_a() {
	if (cond) {
		return {}; //clearly empty
	} else {
		return {value};
	}
}
```
as opposed to:
```
std::optional<T> func_b() {
	if (cond) {
		return std::nullopt; //what is this...? let me see the docs
	} else {
		return {value};
	}
}
```
Emptying an optional can be done by calling `TakeValue()`.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
+#pragma once
+
+#include "BTreeNode.hpp"
+#include "Optional.hpp"
+
+namespace Poly {
+
+	template<typename K, typename V, size_t B>
+	struct Range { //note(vuko): doubles as a proper fucking iterator; behold the goddamn beauty! (alright, this is for show only for now)
```
VukoDrakkeinen: This is a temporary code that is supposed to show off the proof-of-concept, which is why I took the liberty to use strong language, despite your stance on the idea ;)
tl;dr: wilco.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+			this->back = BTree::last_leaf_edge(KVERef{kv_ref.node_ref, kv_ref.idx - 1}.descend());
+			return {kv_ref.node_ref.node->keys[kv_ref.idx], kv_ref.node_ref.node->values[kv_ref.idx]};
+		}
+
+		KV next() {
+			assert(!(front == back));
+			return next_unchecked();
+		}
+	};
```
VukoDrakkeinen: I'd rather have it in a separate PR. Doing it that way will allow me to iron out the kinks in design and won't delay the map any longer. As a side note, a specialized map range can be implemented way more efficiently than the generic one implemented in terms of iterators. In fact for B-trees the iterators themselves are more elegantly implemented in terms of a range encompassing the whole collection and not the other way around!

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+	//todo(vuko): docs
+	template<typename K, typename V, size_t Bfactor = 6>
+	class BTreeMap {
```
VukoDrakkeinen: Eh, `Map` is a bit too undescriptive. I mean, is a `HashMap` not a `Map`? ;) Then there's the issue of how they're different. I was initially thinking in the lines of `HashMap` and `TreeMap`, 
and since `Tree` is literally one character away from `BTree`, here we are. It's good to be clear about things. Well, I could settle for `SortedMap`/`UnorderedMap`, for example.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		using NodeRef  = typename BTree::NodeRef;
+		class Entry;
+		class Iter;
+		class Keys;
+		class Values;
+	public:
+		static constexpr size_t B = Bfactor;
+
+	public:
+		BTreeMap() : root{new LeafNode(), 0}, len(0) {} //todo(vuko): should we delay the allocation until the first insertion?
```
VukoDrakkeinen: Wilco.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		BTreeMap(BTreeMap&& other) : root(other.root), len(other.len) { ::new(&other) BTreeMap(); } //note(vuko): due to the previous issue this allocates!
+		//todo(vuko): copy constructor? tbh I'd rather avoid implementing it
```
VukoDrakkeinen: I will implement the naive algorithm for now.

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		BTreeMap& operator=(BTreeMap&& other) {
+			this->~BTreeMap();
+			this->root = other.root;
+			this->len = other.len;
+			::new(&other) BTreeMap();
+			return *this;
+		}
+
+	public:
+		Optional<V> Insert(const K&  key, const V&  value) { return this->insert(          key ,           value ); }
```
VukoDrakkeinen: What is wrong with `this->`? (pun not intended)

>rename to InsertImpl

Pimples, yuck...
I agree that the public API needs to be consistent and feel that lowercase communicates that the inspected method is not a part it. If that's not acceptable, how
about `ForwardingInsert`?

>What is in the returned Optional<V>?

`Insert()` returns an empty Optional in case the key was not present in the map. Otherwise, the value is updated and the old value is returned. Note that the key is not updated (may matter for types, that can be equal without being identical).

>This whole api needs documentation

I want to write one, once the API is stabilized. In the meantime, one can infer most notions about intended usage and the API details from the unit tests code ;)

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+		Optional<V> Insert(const K&  key,       V&& value) { return this->insert(          key , std::move(value)); }
+		Optional<V> Insert(      K&& key, const V&  value) { return this->insert(std::move(key),           value ); }
+		Optional<V> Insert(      K&& key,       V&& value) { return this->insert(std::move(key), std::move(value)); }
```
VukoDrakkeinen: This can be done using `BTreeMap::Entry`
```
if (auto entry = map.Entry(key)) {
	//examples:
	some_ref = entry.OccupiedGet();
	entry.OccupiedInsert(value_a); //in case we want to replace
} else {
	entry.VacantInsert(value_b); //in case want only insertions
}
```
Now that I think about it, `OccupiedReplace()` seems like a better name... Thoughts?

PolyEngine/Core/Src/BTreeMap.hpp
```diff
@@ -0,0 +1,532 @@
...
+
+		Optional<V> Remove(const K&  key) { return this->remove(key); }
```
VukoDrakkeinen: >returns the value at the key

Yes, you're correct.

>two separate methods

Hmm... why?

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Not sure how I missed this. :+1:

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Wow, that's a textbook elementary case... How did I miss _that_?!

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: I did some thinking on the matter and came to the conclusion that it makes sense to have both iterator and Entry, as they have fairly different use-cases:
##
Entry:
- :ballot_box_with_check: can insert/remove
- :ballot_box_with_check: can update the value
- :no_entry_sign: cannot traverse
- :ballot_box_with_check: can check for existence

##
Iterator:
- :no_entry_sign: cannot insert/move
- :ballot_box_with_check: can update the value 
- :ballot_box_with_check: can traverse
- :heavy_minus_sign: (_does not apply_)

As you can see, there is only minor overlap.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: >Provide const overloads for these six methods

Oh god, this is going to bloat the implementation even more. So much boilerplate... ;_;

>Keys() and Values() should return propper iterator ranges. Ofcourse that implies implementation of KeyIterator and ValueIterator.

I'm not sure what you mean exactly? `Keys` and `Values` are already ranges that can be iterated upon using `begin()` and `end()`.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Done. I previously assumed that, the same problem that we've encountered with component iterators, will be also present here, but it actually works. I wonder if anything can be done about `begin()` and `end()`...

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Methods called without `this->` confuse the hell out of me as I think them to be non-member functions at first :( I think it's better to be explicit, but I will remove these if you insist.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: I agree the API could be nicer if we had:
- sum types *with* pattern matching
- affine type system

The first point would address the possibility of calling wrong method on the wrong variant (e.g. `OccupiedGet()` on `VacantEntry`).
The second would take care of the invalidation problem (as entries would get consumed on insert/remove).
With that being said, it's still not _that_ unsafe. The incorrect variant issue is fully defined and safe (just as `Remove` of `Remove`/`TryRemove` combo is); the assertion will fail.
```
collection.TryRemove(value); //always ok
collection.Remove(value); //may be a programmer error
```
```
map.Entry(of_key).OccupiedGet(); //essentially the same
```
(I now realized that I had forgotten to make entries implicitly convertible to boolean values, which would enable the following pattern ⟨as there are only 2 possible variants⟩)
```
if (auto entry = map.Entry(key)) {
	//do stuff with occupied entry
} else {
	//insert
}
```
And then in C++17 one can do the more descriptive
```
if (auto entry = map.Entry(key); !entry.IsVacant())
```
There is nothing however that can be done about invalidation as evidenced by even the standard library iterators (we need proper lifetime enforcement for that)
```
auto entry = map.Entry(key);
entry.Remove();
entry.Remove(); //uh oh
```
```
auto iter = map.find(foo);
map.insert(bar);
*iter; //oh boy
```

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: >Rename

Wilco.
>Provide ConstIterator

Oh boy, more boilerplate... I wonder if I should move those associated classes out of line to help the readability.

PolyEngine/Core/Src/Optional.hpp
MuniuDev: Ok, I can understand that. Let's keep it just to inlined methods or method declarations only, ok?

PolyEngine/Core/Src/Optional.hpp
MuniuDev: As of right now they don't. The same can be told about ASSERTE macro or custom allocators. But using these wrappers gives us powerfull controll over object lifetime management.

PolyEngine/Core/Src/Optional.hpp
MuniuDev: Let's remove them then ;)

PolyEngine/Core/Src/Optional.hpp
MuniuDev: You can use `Optional` in many other situations, i.e.: for fields in configuration objects, to specify lack of value, or value to use. `Optional<T>` copy constructor and copy operator don't hurt us more than `T` copy constructor. I don't see the point of this limitation.

PolyEngine/Core/Src/Optional.hpp
MuniuDev: StealValue maybe? I can live with take, just seems ackward as the full sentence is: take OUT value.

PolyEngine/Core/Src/Optional.hpp
VukoDrakkeinen: Ok.

PolyEngine/Core/Src/Optional.hpp
MuniuDev: I know i just contradicted myself with previous comments. (about being used for return statements). Now I remembered other use cases.

PolyEngine/Core/Src/Optional.hpp
VukoDrakkeinen: Steal is even weirder, haha. I mean, the more natural way to say it is "take (the) value out". If that doesn't work there is still "take (the) value for ourselves" ;)

PolyEngine/Core/Src/Optional.hpp
MuniuDev: Ok, didn't though about the braces.

PolyEngine/Core/Src/Optional.hpp
VukoDrakkeinen: I'll try to keep it to minimum and only where it makes sense. If I overuse it, shoot a comment.

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: Let's make the description inform us about function not form. What about `OrderedMap` and `UnorderedMap`?

PolyEngine/Core/Src/Optional.hpp
VukoDrakkeinen: Makes sense!

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: `this->` is just against our coding standards. Whats wrong with `InsertImpl` it's far more descriptive thatn `ForwardingInsert`. `Impl` is also widely used in many different libraries and frameworks.

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: That's true, but I want Entry to be replaced by iterators. And it's far more constistent to provide whole map API in the `Map` class itself.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Yeah, `Unordered`/`Ordered` sounds great and should prevent any confusion (although personally I wish for names that would succinctly communicate that `UnorderedMap` is generally faster and to be preferred)

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: For error catching. There are to cases:
1. I don't know anything about the map but I want it to not have entry under `key`
2. I know my key/value is present in the map and I want to remove it. In which case the lack of the key/value should cause an assertion. I used this pattern before and it works very well.

MuniuDev: The same can be said about Insert

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: This needs live discussion on chat. I will reach you.

PolyEngine/Core/Src/BTreeMap.hpp
MuniuDev: We will discuss whole concept of reorganizing Entry, Iterator and Keys/Values in chat.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: >against our coding standards

Alright, I will remove them. Now that I think about it, maybe we could use a formalized guidelines document, that I could refer to, when in doubt.

>Whats wrong with `InsertImpl`
Nothing wrong, I just think that it's ugly as sin and prefer other solutions when possible.

> `InsertImpl` it's far more descriptive thatn `ForwardingInsert`

PolyEngine/Core/Src/Optional.hpp
VukoDrakkeinen: Eh, both your examples do actually provide value on their own :P The argument about debugging/profiling _almost_ convinces me, though I'm left wondering if we have any resources to write such tools that would be better than (or at least as good as) what is already available.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Both cases can be handled by the Entry API and even without that one can just check the returned optional.

PolyEngine/Core/Src/BTreeMap.hpp
VukoDrakkeinen: Alright, I look forward to it.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+
+#include "Defines.hpp"
+#include "UnsafeStorage.hpp"
+
+namespace Poly {
+	namespace Impl {
+
+		//note(vuko): used to move into uninitialized memory, when using operator=() and (by extension) std::move([range]) is not possible
+		template<typename InIt, typename OutIt>
+		auto constructing_move(InIt first, InIt last, OutIt dest_first) {
```
MuniuDev: Change funtion name to PascalCase

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+			for (; first < last; ++first, ++dest_first) {
+				::new(&*(dest_first)) OutType(std::move(*first));
+				first->~InType();
+			}
+
+			return dest_first;
+		};
+
+		template<typename InputIter, typename DestIter>
+		auto constructing_move_backwards(InputIter first, InputIter last, DestIter dest_last) {
```
MuniuDev: Change funtion name to PascalCase

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+
+			struct BranchNode;
+			struct LeafNode {
+				LeafNode() : parent(nullptr), len(0) {}
+				~LeafNode() {
+					//since the backing storage has no idea about how many values it contains, we need to destroy them explicitly
+					keys  .destruct(len);
+					values.destruct(len);
+				}
+				auto as_branch() { return static_cast<BranchNode*>(this); };
```
MuniuDev: Change method name to PascalCase. Do it for whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+			constexpr static size_t CAPACITY = 2 * B - 1u;
+			constexpr static size_t MIN_LEN = B - 1u;
+			static_assert(CAPACITY >= 3, "Capacity must be at least 3 or greater!");
+
+			struct BranchNode;
+			struct LeafNode {
+				LeafNode() : parent(nullptr), len(0) {}
+				~LeafNode() {
+					//since the backing storage has no idea about how many values it contains, we need to destroy them explicitly
+					keys  .destruct(len);
```
MuniuDev: We agreed to use this formatting in method/functions names. Not in method calls.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+					values.destruct(len);
+				}
+				auto as_branch() { return static_cast<BranchNode*>(this); };
+
+				UnsafeStorage<K, CAPACITY> keys;
```
MuniuDev: Member value shoud be names with PascalCase. This comment applies to the whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+			while (first < last) {
+				::new(&*(--dest_last)) OutType(std::move(*(--last)));
+				last->~InType();
+			}
+
+			return dest_last;
+		};
+
+		template<typename K, typename V, size_t B>
+		struct BTree {
```
MuniuDev: I just noticed it. ALL structures and classes need to inherit from BaseObject or (BaseObjectLiteralType in very specific cases). This is major and applies to the whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+				LeafNode* node;
+				size_t height;
+
+				NodeRef as_node_ref() { return {this->height, this->node, this}; };
+
+				NodeRef push_level() {
+					const auto new_node = new BranchNode();
+					new_node->edges[0] = this->node;
+
+					this->node = new_node;
```
MuniuDev: Remove `this->` calls. this applies to whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+
+				template<typename Key, typename Val>
+				void push_back(Key&& key, Val&& value, const Edge edge) {
+					//branch
+					ASSERTE(edge.height == height - 1u, "Attempting to push an edge node of incorrect height!");
+					ASSERTE(node->len < CAPACITY, "No space left in the branch!");
+
+					const size_t idx = node->len;
+
+					::new(&node->keys  [idx]) K(std::forward<Key>(key));
```
MuniuDev: Use ObjectLifetimeHelpers. This applies to whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+					}
+				}
+
+				struct PopResult {
+					K key;
+					V value;
+					Edge edge; //optional
+				};
+
+				PopResult pop_back() {
```
MuniuDev: Braces should always be put in separate line. This applies to whole PR.

PolyEngine/Core/Src/BTreeNode.hpp
```diff
@@ -0,0 +1,605 @@
...
+					ASSERTE(node->len > 0, "No entry to pop in the node!");
+
+					const uint16_t idx = node->len - 1u;
```
MuniuDev: Is there a reason for using uint16 and not normal unsigned or size_t?

PolyEngine/Core/Src/OrderedMap.hpp
```diff
@@ -0,0 +1,696 @@
...
+			/**
+			 * <summary>Removes the entry from the map.</summary>
+			 * <returns>Removed value.</returns>
+			 * <seealso cref="RemoveKV"/>
+			 */
+			V Remove() { return RemoveKV().value; }
+
+		private:
+			template<typename Val>
+			V& or_insert(Val&& value) {
```
MuniuDev: Use PascalCase. If there is naming conflict use `Impl`.

PolyEngine/Core/Src/BTreeNode.hpp
VukoDrakkeinen: Performance/space overhead. It's quite unlikely you're going to need more than 65535 elements in one node (default is 11 :P) and uint16_t fits in nicely.

PR #68, opened 2017-08-24
Fix Release compilation and SGJGame leaking levels

VukoDrakkeinen:


VukoDrakkeinen:
Well, I did not expect MSVC's two-phase lookup implementation to be so ridiculously broken :P You've seen the workaround I came up with and it's not a pretty sight. I need to experiment some more and come up with a better, possibly generic, solution.

MuniuDev:
Oh, ok, I didn't know that this was the issue here. Then this PR can wait for better solution for this problem.

MuniuDev:
Outdated.

PR #70, opened 2017-08-24
Enable compiler cache on Travis

VukoDrakkeinen:


.travis.yml
```diff
@@ -8,6 +8,9 @@ notifications:
 # Enable C++ support
 language: cpp
 
+# Enable compiler cache to speed up compilation
+cache: ccache
```
MuniuDev: If we're enabling cache, then let's ad `make clean` to build command.

VukoDrakkeinen: Only object files are cached. The build files are generated from zero on every build, and so `make clean` wouldn't do anything.

PR #74, opened 2017-10-26
Quit function implementation

Astrael1:


PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -106,3 +106,10 @@ void Engine::ResizeScreen(const ScreenSize & size)
 	gConsole.LogDebug("Screen resize: {} {}", size.Width, size.Height);
 	GetRenderingDevice()->Resize(size);
 }
+
+void Engine::QuitGame()
+{
+	shouldQuit = 1;
```
MuniuDev: Please nevere use numbers for boolean value assignments. Use `true` and `false` instead. Also this implementation can be written in one line.

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -123,6 +123,11 @@ namespace Poly
 		/// <param name="pos">New wheel position.</param>
 		void UpdateWheelPos(const Vector& pos) { InputEventsQueue.PushBack({eInputEventType::WHEELMOVE, pos}); }
 
+		//functions for closing the game
+		bool ShouldQuit();
```
MuniuDev: `ShouldQuit` should be implemented as const method (`void ShouldQuit() const;`). Also please provide XML documentation for this method describing how is it used. I also suggest rename to `IsGameQuiteRequested`.

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -123,6 +123,11 @@ namespace Poly
...
+		void QuitGame();
```
MuniuDev: Let's rename it to `RequestGameQuit`. Also please provide XML documentation (simmilar to other methods in this file).

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -223,6 +228,8 @@ namespace Poly
 
 		std::unordered_map<std::type_index, size_t> ComponentTypeMap;
 		std::unordered_map<std::type_index, size_t> WorldComponentTypeMap;
+
+		bool shouldQuit = 0; //stop the game
```
MuniuDev: According to our coding style members should be names using PascalCase, not cameCase. Also let's rename it to `GameQuitRequested` and initialize with `false` and not 0.

PolyEngine/Standalone/Windows/Src/Main.cpp
```diff
@@ -38,6 +38,8 @@ Poly::IGame* LoadGame()
 	return getGame();
 }
 
+
+
```
MuniuDev: Remove these unnecessary lines.

PolyEngine/Standalone/Windows/Src/Main.cpp
```diff
@@ -218,3 +221,5 @@ LRESULT CALLBACK WindowProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lPara
 	// Handle any messages the switch statement didn't
 	return DefWindowProc(hWnd, message, wParam, lParam);
 }
+
+
```
MuniuDev: Remove these unnecessary lines.

PolyEngine/Standalone/Windows/Src/Main.cpp
```diff
@@ -126,6 +128,7 @@ int WINAPI WinMain(HINSTANCE hInstance,
 				quitRequested = true;
 		}
 
+		quitRequested |= Engine.ShouldQuit();
```
MuniuDev: `|=` performs bitwise or and assign operation which is invalid in this context. Please replace it with `quitRequested = quitRequested || Engine.ShouldQuit();`. Also let's move this implementation after call to `Update`.

PolyEngine/Engine/Src/Engine.hpp
VukoDrakkeinen: Hm, I don't know... `RequestGameQuit()`/`IsGameQuitRequested()` is grammatically incorrect and sounds like quitting is optional and can be denied.
How about `QuitGame()`/`GameAboutToQuit()` or `QuitGame()`/`IsGameQuitting()`?

VukoDrakkeinen: Apparently _quit_ as a noun is a collective term for some types of bird species, as I was surprised to find out 😛

PolyEngine/Standalone/Windows/Src/Main.cpp
VukoDrakkeinen: One wonders why there is no corresponding `||=` operator, eh? 😉

PR #78, opened 2017-11-01
[Rendering] phong shading

sp0lsh:


sp0lsh:
Now I see that this is utterly old branch, not going to add fixes there.
Will add Phong Feature changes in new fresh brand new branch and then create PR.

PR #81, opened 2017-11-02
Switch to compile time component indexing

filu005:
Redone component indexing; Run time typeid() was replaced by compile time indexing (based on templates).
Removed component registration.
Every component used in two different compilation units must(!) have appropriate template instantiation. For details see Family.hpp.


PolyEngine/Engine/Src/DebugWorldComponent.hpp
```diff
@@ -17,4 +17,5 @@ namespace Poly
 	private:
 		FPSSystem::FPSData FPSData;
 	};
+	COMPONENT_INS(WorldComponentsFamily, DebugWorldComponent)
```
MuniuDev: Let's rename this macro to REGISTER_COMPONENT.

PolyEngine/Engine/Src/Family.hpp
```diff
@@ -0,0 +1,59 @@
...
+	/// <para>
+	/// Definitions are kept in Family.inc file. That means that this '.inc' file must
+	/// be included in every .cpp file where template of Family with specific component type
+	/// is used for the first time. This allows compiler to create a definition
+	/// of Family::type_id<U>() type (where U is typeof the component).
+	/// </para>
+	template<typename T>
+	class Family
+	{
+		static size_t identifier();
```
MuniuDev: Use PascalCase for methods.

PolyEngine/Engine/Src/Family.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+	public:
+		///<summary>
+		/// Returns an unique identifier for the given type.
+		///</summary>
+		template<typename U>
+		static size_t type_id();
```
MuniuDev: `type_id` is a very bad method name. It resebles `typeid` which is specific feature of C++. Rename it, use PascalCase.

PolyEngine/Engine/Src/Family.inc
```diff
@@ -0,0 +1,35 @@
...
+	{
+		static const size_t value = identifier();
+		return value;
+	}
+
+	EXPIMP_TEMPLATE template class
+#if (defined(_GAME) && (defined(__GNUC__) || defined(__clang__)))
+
+#else
+	ENGINE_DLLEXPORT
```
MuniuDev: ENGINE_DLLEXPORT is not defined when compiling for *nix anyway. This ifdef is pointless.

PolyEngine/Engine/Src/Family.inc
```diff
@@ -0,0 +1,35 @@
...
+#endif
+	Family<struct ComponentsFamilyType>;
+
+	EXPIMP_TEMPLATE template class
+#if (defined(_GAME) && (defined(__GNUC__) || defined(__clang__)))
+
+#else
+	ENGINE_DLLEXPORT
```
MuniuDev: ENGINE_DLLEXPORT is not defined when compiling for *nix anyway. This ifdef is pointless.

PolyEngine/Engine/Src/World.hpp
```diff
@@ -224,8 +230,9 @@ namespace Poly {
 	template<typename T>
 	T* Entity::GetComponent()
 	{
-		if (HasComponent(gEngine->GetComponentID<T>()))
-			return static_cast<T*>(Components[gEngine->GetComponentID<T>()]);
+		auto const ctypeID = ComponentsFamily::type_id<T>();
```
MuniuDev: change to `const auto`, do it in all occurences.

PolyEngine/Games/SandboxRendering/Src/RenderingSandbox.cpp
```diff
@@ -15,6 +15,8 @@
 #include <TimeWorldComponent.hpp>
 #include <SoundSystem.hpp>
 
+#include "Family.inc"
```
MuniuDev: These includes are kind of dealbreaker for this whole feature. Maybe we should reconsider runtime id generation?

PolyEngine/Engine/Src/Engine.hpp
```diff
@@ -127,60 +87,6 @@ namespace Poly
...
-				})(),
-				"Component type or id was registered twice!");
-
-			ComponentTypeMap[typeid(T)] = id;
-		}
-
-		/// <summary>If given component is registered function returns associated ID.</summary>
-		/// <tparam name="T">Typename which ID is requested</tparam>
-		/// <returns>Associated ID.</returns>
-		template<typename T> size_t GetComponentID() const
```
MuniuDev: It would be nicer to use this method instead of the one in Family for getting component ID. The we can hide Family behind `Impl` namespace.

PolyEngine/Engine/Src/Family.hpp
```diff
@@ -0,0 +1,59 @@
...
+	/// </para>
+
+	/// <para>
+	/// Definitions are kept in Family.inc file. That means that this '.inc' file must
+	/// be included in every .cpp file where template of Family with specific component type
+	/// is used for the first time. This allows compiler to create a definition
+	/// of Family::type_id<U>() type (where U is typeof the component).
+	/// </para>
+	template<typename T>
+	class Family
```
MuniuDev: We should come up with better name than Family.

PolyEngine/Engine/Src/Engine.hpp
filu005: Moving this method from Engine.hpp to World.hpp saves us one pointer call (i.e. abandon referencing via gEngine-> in World.hpp).
Apart from that, in World.hpp we create the division into World/'Regular' components (AddComponent(), AddWorldComponent()) so the division of Family naturally fits in there.

PolyEngine/Engine/Src/Family.hpp
filu005: It kinda comes from biology (https://en.wikipedia.org/wiki/Family_(biology)) where Family can be divided into Genera (ComponentsFamily, WorldComponentsFamily) and Genus breakes down into Species (CameraComponent, InputWorldComponent).
But I agree, it's not a great name; it isn't descriptive.

PolyEngine/Engine/Src/Family.inc
filu005: Under GCC and clang this macro always turns into: `__attribute__ ((visibility ("default")))`. This attribute causes a compilation error in `_GAME` projects.
IMO this empty `#if` describes what I want to achieve: in `_GAME` under GCC and clang leave it empty. AFAIK this attribute error might be a bug (in GCC) so I want here to highlight that.

PolyEngine/Engine/CMakeLists.txt
```diff
@@ -64,6 +65,8 @@ set(POLYENGINE_H_FOR_IDE
 	Src/Engine.hpp
 	Src/EnginePCH.hpp
 	Src/Entity.hpp
+	Src/Family.hpp
+	Src/Family.inc
```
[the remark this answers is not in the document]
filu005: done

PolyEngine/Engine/Src/DebugWorldComponent.hpp
filu005: done

PolyEngine/Engine/Src/Family.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#if defined(_ENGINE)
+#    define EXPIMP_TEMPLATE
+#else
+#    define EXPIMP_TEMPLATE extern
+#endif
+
+#define COMPONENT_INS(FAMILY, COMPONENT) \
+	EXPIMP_TEMPLATE template size_t ENGINE_DLLEXPORT FAMILY::type_id<COMPONENT>();
```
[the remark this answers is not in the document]
filu005: sure! done!

PolyEngine/Engine/Src/Family.hpp
filu005: done

PolyEngine/Engine/Src/Family.hpp
filu005: replaced with GetComponentTypeID()

PolyEngine/Engine/Src/World.hpp
filu005: done

PolyEngine/Games/SandboxRendering/Src/RenderingSandbox.cpp
filu005: I've figured out that includes in .cpp files are not necessary.
Definitions are now included once in ComponentBase.hpp.
problem solved!

PolyEngine/Engine/Src/Engine.hpp
filu005: done. created static method GetComponentID() in World.hpp

PolyEngine/Engine/Src/Family.hpp
filu005: done. renamed to ComponentIDGenerator

PR #82, opened 2017-11-03
Enum array initialization extension.

PiotrMoscicki:


PolyEngine/Core/Src/EnumUtils.hpp
```diff
@@ -68,6 +68,12 @@ namespace Poly {
 		EnumArray() : Data{} {}
 
 		//------------------------------------------------------------------------------
+		EnumArray(const std::initializer_list<std::pair<E, T>>& list)
```
MuniuDev: Please also provide implementation for `operator=`

PolyEngine/Core/Src/EnumUtils.hpp
```diff
@@ -120,6 +126,13 @@ namespace Poly {
...
 
 		//------------------------------------------------------------------------------
+		void PopulateFromInitializerList(const std::initializer_list<std::pair<E, T>>& list)
+		{
+			for (const std::pair<E, T>& val : list)
+				Data[static_cast<int>(val.first)] = val.second;
+		}
+
+		//------------------------------------------------------------------------------
 		void PopulateFromInitializerList(const std::initializer_list<T>& list)
```
MuniuDev: Let's remove creation from `std::initializer_list<T>` alltogether and either force to declare enum names by providing both the enum value and enum name or refactor enum names declaration to create `std::initializer_list<E,T>` on the fly. Preferably both.

PolyEngine/Core/Src/EnumUtils.hpp
vadordev: I only took a glimpse at it but what about an rvalue reference? 
To support std::move( T ) etc.

vadordev: If that is possible, that is. I don't know initializer_list that well.

MuniuDev: Thats actually a good idea. We would have to move each element separately though.

vadordev: How does one do that?
Variadic templates?

VukoDrakkeinen: Nananana, C++17-compliant Vuko to the rescue! 😜 
According to the standard, "_copying a std::initializer_list does not copy the underlying objects_" so we could just pass it by value. To add to that, the elements themselves are copy-initialized and then copied out of the underlying array. Unfortunately, moving them out is impossible since the array may be located in read-only memory.

PR #83, opened 2017-11-05
config serialization WIP

MuniuDev:


PolyEngine/Engine/Src/ConfigBase.cpp
```diff
@@ -1,18 +1,86 @@
...
+	// For now, ensure newest state of config file.
+	Save(); 
 }
 
-void Poly::ConfigBase::Load()
+const String& ConfigBase::GetFileName() const
 {
+	if (FileName.GetLength() == 0)
+		FileName = String("Configs/") + String(GetTypeInfo().GetTypeName()) + String(".json");
+	return FileName;
```
vadordev: Why not make _FileName_ a static variable inside of this function, since it is not used anywhere outside of this getter anyway?
Instead of making the hack with mutable in _.hpp_ file?

MuniuDev: Because FileName for different configs will be different.

VukoDrakkeinen: I wonder if using a _curiously recurring template pattern_ here would be an overkill...
i.e. `DebugConfig : ConfigBase<DebugConfig>` which would yield a unique static variable for each derived config.

MuniuDev: This is totally unnecessary. Variable works just fine.

VukoDrakkeinen: Yeah, I was just throwing ideas around for the sake of it. Caching `mutable` variable seems fine to me (simpler with the same performance characteristic).

PolyEngine/Engine/Src/ConfigBase.hpp
```diff
@@ -9,7 +9,8 @@ namespace Poly
 	{
 		ENGINE,
 		GAME,
-		LOCAL
+		LOCAL,
+		RUNTIME_DIR
 	};
```
VukoDrakkeinen: We could use some comments here 😛 I'm assuming `RUNTIME_DIR` is `.`, `ENGINE` is some engine-specific config location (why do we need that?), while `GAME` is accordingly game-specific... What's `LOCAL`?

MuniuDev: True,
ENGINE will contain all engine configuration
GAME contains all game configs
LOCAL is all user configs stored in platform specific location
RUNTIME_DIR is what you said.

This enum will be probably merged with eResourceSource

VukoDrakkeinen: So why do we need separate directories for engine and game configuration files?
How is `LOCAL` different from `ENGINE` and `GAME`? Is it a parent directory? If so why do we need an explicit enumeration for it?

PolyEngine/Engine/Src/AssetsPathConfig.cpp
```diff
@@ -1,140 +1,35 @@
...
-
+		case eResourceSource::ENGINE:
+			return EngineAssetsPath;
+		case eResourceSource::GAME:
+			return GameAssetsPath;
+		case eResourceSource::NONE:
+		case eResourceSource::USR_LOCAL: //TODO implement usr local directory
+			return EMPTY;
+		default:
+			ASSERTE(false, "Invalid resoruce source!");
```
VukoDrakkeinen: Type-oh

PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -11,6 +11,9 @@ Engine::Engine()
 {
 	ASSERTE(gEngine == nullptr, "Creating engine twice?");
 	gEngine = this;
+
+	gAssetsPathConfig.Load();
+	gDebugConfig.Load();
```
VukoDrakkeinen: I'm not sure about having to remember to manually load the configuration, given that you get the defaults if you forget to. Sounds like a future common source of silent config-overwriting bugs.

PolyEngine/Engine/Src/AssetsPathConfig.cpp
```diff
@@ -1,140 +1,35 @@
...
-		rapidjson::Document Json; // UTF8 by default
-		if (Json.Parse(ConfigFileContent.GetCStr()).HasParseError())
-		{
-			IsSuccessfull = false;
-			gConsole.LogInfo("Config file json has error: {}", Json.GetParseError());
-
-			LoadDefaults();
-		}
-		else
+		static String EMPTY;
```
VukoDrakkeinen: Why is this static and not a global constant?

PolyEngine/UnitTests/Src/ConfigTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+	eConfigTest PropEnum = eConfigTest::VAL_2;
+};
+RTTI_DEFINE_TYPE(TestConfig)
+
+TEST_CASE("Config serialization tests", "[ConfigBase]")
+{
+	TestConfig config;
+	config.Save();
+	config.Load();
+}
```
VukoDrakkeinen: You might have forgotten about something here, haha

PolyEngine/Engine/Src/ResourceManager.cpp
```diff
@@ -14,26 +14,29 @@ DEFINE_RESOURCE(TextureResource, gTextureResourcesMap)
...
 
-ENGINE_DLLEXPORT String Poly::LoadTextFileRelative(eResourceSource Source, const String & path)
+String Poly::LoadTextFileRelative(eResourceSource Source, const String & path)
 {
 	bool IsNotLoaded = true;
 
 	String FileContent;
-	Dynarray<String> Paths = (Source == eResourceSource::NONE) ? Dynarray<String>{""} : gAssetsPathConfig.GetAssetsPaths(Source);
-	for (size_t i = 0; i < Paths.GetSize() && IsNotLoaded; ++i)
+	String AbsolutePath = gAssetsPathConfig.GetAssetsPath(Source) + path;
```
VukoDrakkeinen: This depends on paths having a trailing separator, which might not always be the case

PolyEngine/Engine/Src/ConfigBase.hpp
```diff
@@ -5,25 +5,29 @@
 
 namespace Poly
 {
-	enum class eConfigLocation
+	/// <summary>Enum used to identify source of asset.
```
VukoDrakkeinen: `s/source of asset/source location of an asset/`

PolyEngine/Engine/Src/ConfigBase.hpp
```diff
@@ -5,25 +5,29 @@
...
+	enum class eResourceSource
 	{
 		ENGINE,
 		GAME,
-		LOCAL
+		NONE,
+		USR_LOCAL,
+		_COUNT
```
VukoDrakkeinen: So why do we need separate directories for engine and game configuration files?
How is `USR_LOCAL` different from `ENGINE` and `GAME`? Is it a parent directory? If so why do we need an explicit enumeration for it?

MuniuDev: We discussed this before. USR_LOCAL is e.g.: %AppData%. Engine has it's own configuration. That is not part of the Game.

PolyEngine/Engine/Src/Engine.cpp
MuniuDev: Well, now this is simply part of initialization process. In the future it would be good to implement some config manager for this. The problem is, some configuration files need to be loaded first.

PolyEngine/Engine/Src/ResourceManager.cpp
MuniuDev: `Path` wrapper is planned. It will fix this kind of problems.

PolyEngine/Core/CMakeLists.txt
```diff
@@ -61,6 +65,7 @@ add_library(polycore SHARED ${POLYCORE_SRCS} ${POLYCORE_H_FOR_IDE})
 target_compile_options(polycore PRIVATE $<$<BOOL:${SIMD}>:${SIMD_FLAGS}>)
 target_compile_definitions(polycore PRIVATE _CORE DISABLE_SIMD=$<NOT:$<BOOL:${SIMD}>>)
 target_include_directories(polycore PUBLIC ${POLYCORE_INCLUDE})
+target_link_libraries(polycore PUBLIC Rapid::JSON)
```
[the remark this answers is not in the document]
MuniuDev: fixed

PolyEngine/Core/Src/RTTIProperty.hpp
```diff
@@ -7,33 +7,101 @@
...
 namespace Poly {
 	namespace RTTI {
 
 		template<typename T, typename U>
 		constexpr size_t OffsetOfMember(U T::*member) { return (char*)&((T*)nullptr->*member) - (char*)nullptr; }
 
+		enum class eCorePropertyType
+		{
+			UNHANDLED,
+			NONE,
```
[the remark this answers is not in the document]
MuniuDev: UNHANDLED is default value, that shows problem in implementation. NONE is a statement that this property is not CorePropertyType (e.g. custom struct)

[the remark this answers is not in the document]
VukoDrakkeinen: A bit "not ideal", yeah 😜 What other types are there that are not core aside from structs?

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+	auto& value = doc.SetObject();
+	RTTI::SerializeObject(obj, propertyName, value, doc.GetAllocator());
+}
+
+void RTTI::SerializeObject(const RTTIBase* obj, const String& propertyName, rapidjson::Value& currentValue, rapidjson::Document::AllocatorType& alloc)
+{
+	const TypeInfo typeInfo = obj->GetTypeInfo();
+	const PropertyManagerBase* propMgr = obj->GetPropertyManager();
+
+	HEAVY_ASSERTE(currentValue.IsObject(), "JSON value is not an object!");
```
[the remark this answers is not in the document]
MuniuDev: not true. This can happen in json DOM that is different from RTTI property tree structure.

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+		const String* str = reinterpret_cast<const String*>(value);
+		currentValue.SetString(str->GetCStr(), alloc);
+		break;
+	}
+	case eCorePropertyType::ENUM:
+	{
+		HEAVY_ASSERTE(prop.ImplData.get() != nullptr, "Invalid enum impl data!");
+		const EnumPropertyImplData* implData = static_cast<const EnumPropertyImplData*>(prop.ImplData.get());
+		i64 val;
+		if (implData->EnumInfo->GetUnderlyingValueSize() == sizeof(i32))
```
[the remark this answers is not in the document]
MuniuDev: In enum serialization we support values that are in range of `i64` type.

PolyEngine/Engine/Src/ConfigBase.hpp
VukoDrakkeinen: Wait, so where do `ENGINE` and `GAME` point to?
>Engine has it's own configuration. That is not part of the Game.

Well, I'm asking what's the reasoning here.

PolyEngine/Engine/Src/Engine.cpp
VukoDrakkeinen: Resolving a dependency tree, hmm... Might be not trivial.

PolyEngine/Engine/Src/ResourceManager.cpp
VukoDrakkeinen: Right. But we do not have it yet and so the code is currently subtly broken.

PolyEngine/Core/Src/RTTISerialization.cpp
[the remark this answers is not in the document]
VukoDrakkeinen: If the API is private, then the DOM is ALWAYS correct (barring bugs, of course).

PolyEngine/Core/Src/RTTISerialization.cpp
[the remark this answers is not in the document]
VukoDrakkeinen: Noted. Could you also answer my other 2 questions, please?

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+	switch (prop.CoreType)
+	{
+	case eCorePropertyType::BOOL:
+		*reinterpret_cast<bool*>(obj) = value.GetBool();
+		break;
+	case eCorePropertyType::INT8:
+	{
+		int result = value.GetInt();
+		HEAVY_ASSERTE((result < 0) ? ((-result & 0xFF) == -result) : ((result & 0xFF) == result), "Value outside of int8 range");
```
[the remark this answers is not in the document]
MuniuDev: This calls for improved assertions, with string format (like in gConsole) which is under developement.

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+	case eCorePropertyType::FLOAT:
+		*reinterpret_cast<float*>(obj) = value.GetFloat();
+		break;
+	case eCorePropertyType::DOUBLE:
+		*reinterpret_cast<double*>(obj) = value.GetDouble();
+		break;
+	case eCorePropertyType::STRING:
+	{
+		String* str = reinterpret_cast<String*>(obj);
+		*str = String(value.GetString());
```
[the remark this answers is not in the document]
MuniuDev: fixed

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+		const EnumPropertyImplData* implData = static_cast<const EnumPropertyImplData*>(prop.ImplData.get());
+
+		const i64 val = implData->EnumInfo->GetEnumValue(value.GetString());
+
+		if (implData->EnumInfo->GetUnderlyingValueSize() == sizeof(i32))
+			*reinterpret_cast<i32*>(obj) = (i32)val;
+		else if (implData->EnumInfo->GetUnderlyingValueSize() == sizeof(i64))
+			*reinterpret_cast<i64*>(obj) = val;
+		else
+			ASSERTE(false, "Unhadled value size!");
```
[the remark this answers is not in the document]
MuniuDev: fixed

PolyEngine/Core/Src/RTTISerialization.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include "RTTI.hpp"
+#include "RTTIProperty.hpp"
+#include <rapidjson/document.h>
+
+namespace Poly
+{
+	namespace RTTI
+	{
+		CORE_DLLEXPORT void SerializeObject(const RTTIBase* obj, const String& propertyName, rapidjson::Document& doc);
+		CORE_DLLEXPORT void SerializeObject(const RTTIBase* obj, const String& propertyName, rapidjson::Value& currentValue, rapidjson::Document::AllocatorType& alloc);
```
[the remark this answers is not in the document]
MuniuDev: I agree this is not ideal. SAX will be considered in the future. When we will consider binary serialization, propper abstraction will be implemented. Right now this is not high priority.

PolyEngine/Engine/Src/AssetsPathConfig.cpp
MuniuDev: good idea!

PolyEngine/Core/Src/RTTIProperty.hpp
```diff
@@ -7,33 +7,101 @@
...
+			eCorePropertyType CoreType;
+			std::shared_ptr<PropertyImplData> ImplData; // @fixme ugly hack for not working Two-phase lookup in MSVC, should use unique_ptr
+		};
+
+		struct EnumPropertyImplData : public PropertyImplData
+		{
+			const ::Poly::Impl::EnumInfoBase* EnumInfo = nullptr;
 		};
 
+		template <typename E> Property CreateEnumPropertyInfo(size_t offset, const char* name, ePropertyFlag flags)
```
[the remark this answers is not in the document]
MuniuDev: implemented constexpr_if and constexpr_match and used it.

PolyEngine/Core/Src/RTTISerialization.cpp
```diff
@@ -0,0 +1,223 @@
...
+CORE_DLLEXPORT void Poly::RTTI::SetCorePropertyValue(void* obj, const RTTI::Property& prop, const rapidjson::Value& value)
+{
+	switch (prop.CoreType)
+	{
+	case eCorePropertyType::BOOL:
+		*reinterpret_cast<bool*>(obj) = value.GetBool();
+		break;
+	case eCorePropertyType::INT8:
+	{
+		int result = value.GetInt();
```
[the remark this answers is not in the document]
MuniuDev: I'm pretty sure rapidjson does all lossless convertions automatically, so in the end we only need to ensure we can fit given value in our object, which we do.

PolyEngine/Engine/Src/AssetsPathConfig.cpp
MuniuDev: fixed

PolyEngine/Core/Src/RTTISerialization.cpp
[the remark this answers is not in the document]
VukoDrakkeinen: Consider this:
```
{
    field: "very large numeric value, for example 2^64"
}
```
We tell rapidjson to parse the field and we want an int: `int result = value.GetInt();`
It does not throw, does not return a sentinel value. What value is `result`?

PolyEngine/Core/Src/RTTISerialization.cpp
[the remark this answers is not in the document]
VukoDrakkeinen: Then we should put a TODO marker here.

PolyEngine/Engine/Src/ConfigBase.hpp
```diff
@@ -5,25 +5,29 @@
...
+	enum class eResourceSource
 	{
-		ENGINE,
-		GAME,
-		LOCAL
+		ENGINE, // engine resources location
+		GAME, // game resources location
+		USR_LOCAL, // user local space resources location
+		NONE, // default (often runtime directory) resources location, this option will not modify path to resource
+		_COUNT
```
VukoDrakkeinen: Copy-pasting my comment, since it got hidden:

VukoDrakkeinen: Wait, so where do `ENGINE` and `GAME` point to?
>Engine has it's own configuration. That is not part of the Game.

Well, I'm asking what's the reasoning here.

MuniuDev:
This change breaks macOS build...

PolyEngine/Engine/Src/ConfigBase.hpp
MuniuDev: e.g. Keeping one configuration of the engine for different projects (games) or acessing shaders taht are part of engine resources.

MuniuDev:
Fuck it.

PR #84, opened 2017-11-07
String api improvements

gkocik:


PolyEngine/Core/Src/String.cpp
```diff
@@ -68,12 +68,16 @@ String String::From(const std::string& var) {
 	return String(var.c_str());
 }
 
-bool String::Contains(const String& /*var*/) const {
-	ASSERTE(false, "Not impleented yet.");
-	//TODO Dynarray.FindAllIdx()
-	//for all this.Data.FindAllIdx()
-	//	if var == this->Substring(idx, idx+var.Data.GetSize()
-	//		return true
+bool String::Contains(const String& var) const {
```
MuniuDev: This implementation is really naive. Much better would be to use two index variables approach. 
`idx1` is incremented for each letter in analyzed string
`idx2` is incremented only when `var[idx2] == Data[idx1]` otherwise it's zeroed.
The string is found when `idx2 == var.GetLength()`. Otherwise string was not found.

PolyEngine/Core/Src/String.cpp
```diff
@@ -126,11 +130,10 @@ String String::Replace(char what, char with) const {
...
 
-String String::Replace(const String& /*what*/, const String& /*with*/) const {
-	ASSERTE(false, "Not implemented yet.");
-	//split source by delimiter a
-	//return string joined with separator b
-	return String();
+String String::Replace(const String& what, const String& with) const {
+	String s = "";
+	Dynarray<String> splitted = this->Split(what);
+	return s.Join(splitted.GetData(), splitted.GetSize(), with);
```
MuniuDev: Join is a static method. Remove variable `s` and simply call `return Join(splitted.GetData(), splitted.GetSize(), with);`

PolyEngine/Core/Src/String.cpp
```diff
@@ -69,12 +69,14 @@ String String::From(const std::string& var) {
...
-		String substr = this->Substring(idx[i], idx[i] + dataSize);
-		if (var == this->Substring(idx[i], idx[i] + dataSize)) {
+	size_t idx1 = 0, idx2 = 0;
+	for (idx1; idx1 < this->GetLength(); idx1++) {
+		if (var[idx2] == Data[idx1]) {
+			idx2++;
+		} else {
+			idx2 = 0;
+		}
+		if (idx2 == var.GetLength()) {
```
MuniuDev: We generally discourage usage of `this->`. You can also get rid of the brackets fo two if statements. Also each bracket should be put in separate line according to our coding style.

VukoDrakkeinen: @MuniuDev: I like how every statement in your comment is in direct opposition to my personal coding style, hahaha.

PR #85, opened 2017-11-07
Rendering: Blinn-Phong pt 2

sp0lsh:
with editable PhongMaterial for MeshRenderingComponents


PolyEngine/Engine/Res/Shaders/blinn-phongFrag.shader
```diff
@@ -44,73 +52,108 @@ in vec3 vNormal;
...
-	float attentuation = 1.0 / (1.0 + light.Attenuation * pow(distanceToLight, 2));
-	return mix(0.0, attentuation, step(distanceToLight, light.Range));
+// returns intensity of specular reflection
+vec3 specularLighting(in vec3 N, in vec3 L, in vec3 V, in vec3 LightColor)
+{
+	float specularTerm = 0;
+
+	// calculate specular reflection only if
+	// the surface is oriented to the light source
+	if (dot(N, L) > 0)
```
MuniuDev: We can get rid of branching here by removing if and changing assigment to `specularTerm` like that:
````
const float dotNL = dot(N, L);
specularTerm = ceil(dotNL * dotNL) * pow(dot(N, H), uMaterial.Shininess);

Comment (MuniuDev) [reply PolyEngine/Engine/Res/Shaders/blinn-phongFrag.shader]: We can do that because, assuming N and L are normal vectors, `ceil(dotNL * dotNL)` will always yield result 0  if `dotNL  == 0` and 1 if `dotNL  != 0`.

File: PolyEngine/Engine/Src/LightSourceComponent.cpp [hunk 2]
```diff
@@ -2,17 +2,17 @@
 
 #include "LightSourceComponent.hpp"
 
-Poly::AmbientLightWorldComponent::AmbientLightWorldComponent(const Color& color, float intensity)
+Poly::AmbientLightWorldComponent::AmbientLightWorldComponent(const Color& color = Color(1.0f, 1.0f, 1.0f), float intensity = 1.0f)
```
MuniuDev: `Color(1.0f, 1.0f, 1.0f)` is used very often, lets provide static fields for each common color, like: `Color::RED`, `Color::BLACK`, etc. Use these fields for initialization.
Similar thing is done in Vector class.

PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
```diff
@@ -19,10 +19,16 @@ const size_t MAX_LIGHT_COUNT_DIRECTIONAL = 8;
...
 
-	GetProgram().RegisterUniform("vec4", "uAmbientLight.Base.Color");
-	GetProgram().RegisterUniform("float", "uAmbientLight.Base.Intensity");
+	GetProgram().RegisterUniform("vec4", "uMaterial.Ambient");
+	GetProgram().RegisterUniform("vec4", "uMaterial.Diffuse");
+	GetProgram().RegisterUniform("vec4", "uMaterial.Specular");
+	GetProgram().RegisterUniform("float", "uMaterial.Shininess");
+
+	GetProgram().RegisterUniform("vec4", "uAmbientLight.Color");
+	GetProgram().RegisterUniform("float", "uAmbientLight.Intensity");
```
MuniuDev: We need to start thinking about adding UBO support.

PolyEngine/RenderingDevice/OpenGL/Src/GLShaderProgram.cpp
```diff
@@ -149,6 +149,10 @@ void GLShaderProgram::SetUniform(const String& name, int val)
 		HEAVY_ASSERTE(it->second.TypeName == "int" || it->second.TypeName == "sampler2D", "Invalid uniform type!");
 		glUniform1i(it->second.Location, val);
 	}
+	// else
+	// {
+	// 	gConsole.LogWarning("GLShaderProgram::SetUniform(): Unregistered Uniform: {}", name);
+	// }
```
MuniuDev: Why are we commenting this out?

sp0lsh: I've spent some time investigating why uniforms in shaders are not updating. It was because uniforms were not registered in program. Mistakes like that should be logged but those lines are called several times each mesh resulting in massive log spam. Logging it only once on first attempt to set unregistered uniform would be appropriate.

PolyEngine/Engine/Src/LightSourceComponent.cpp
sp0lsh: Done

PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
sp0lsh: Will start think about UBO, but I would do Forward+ first to have clear picture on which uniforms shading do exactly need.

PolyEngine/Engine/Res/Shaders/blinn-phongFrag.shader
sp0lsh: Good idea! Since we care only about positive range of `dot(N, L)` it can be just `ceil(dot(N, L))`:
`specularTerm = ceil(dot(N,L)) * pow(dot(N, H), uMaterial.Shininess);`

checked it with: https://www.shadertoy.com/view/MdjGRy
in GLSL `ceil(x)` on `[-1, 0]` is 0
and specular looks the same on render

PolyEngine/Core/Src/Color.hpp
```diff
@@ -12,6 +12,12 @@ namespace Poly {
 		Color();
 		Color(float r, float g, float b, float a = 1.0f);
 
+		// Operators with other Colors
+		Color operator+(const Color& rhs) const;
+		Color operator-(const Color& rhs) const;
+		Color& operator+=(const Color& rhs);
+		Color& operator-=(const Color& rhs);
```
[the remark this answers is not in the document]
sp0lsh: Did them analogy to Vector. Now moved to header. Added my first constexpr (required to change BaseObject to BaseObjectLiteralType) and it even compiled, oh boy! Enough adrenaline for today ;)

PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
sp0lsh: UBO from Unreal Engine 4:
https://docs.unrealengine.com/latest/INT/API/Runtime/Engine/FViewUniformShaderParameters/index.html

PR #87, opened 2017-11-08
[WiP] SafePointer

DevRayy:
This is still work in progress, PRing now to get some feedback.


PolyEngine/Core/Src/SafePtr.hpp
```diff
@@ -0,0 +1,29 @@
...
+	template<typename T>
+	class SafePtr : public BaseObject<>
+	{
+	public:
+		SafePtr(SafePtrRoot* pointer)
+		{
+			PointerIdx = SafePtrRoot::RegisterPointer(pointer);
+		}
+
+		SafePtrRoot *operator->()
```
MuniuDev: This operator should return `T*`. There should also be a const version.

PolyEngine/Core/Src/SafePtr.hpp
```diff
@@ -0,0 +1,29 @@
...
+		{
+			return SafePtrRoot::GetPointer(PointerIdx);
+		}
+
+		SafePtrRoot *operator*()
```
MuniuDev: This operator should return `T*`. There should also be a const version.

PolyEngine/Core/Src/SafePtr.hpp
```diff
@@ -0,0 +1,29 @@
...
+		{
+			return SafePtrRoot::GetPointer(PointerIdx);
+		}
+	private:
```
MuniuDev: This API is not complete. We need eq and neq comparison operators (with `SafePtr<T>` and with `T*`).  Also Get() method that returns `*T` alongside with const version.

PolyEngine/Core/Src/SafePtrRoot.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include "Defines.hpp"
+#include "Dynarray.hpp"
+
+namespace Poly {
+	class CORE_DLLEXPORT SafePtrRoot : public BaseObject<>
+	{
+	public:
+		//TODO remove later
+		int Debug();
```
MuniuDev: Then let's remove it now :)

PolyEngine/Core/Src/SafePtrRoot.cpp
```diff
@@ -0,0 +1,46 @@
...
+//------------------------------------------------------------------------------
+SafePtrRoot::~SafePtrRoot()
+{
+	SafePtrRoot::ClearPointer(this);
+}
+
+//------------------------------------------------------------------------------
+size_t SafePtrRoot::RegisterPointer(SafePtrRoot *pointer)
+{
+	SafePtrRoot::Pointers.PushBack(pointer);
```
MuniuDev: This method should check if the pointer was already registered once. 
```  
  const auto it = PointersMap.find(pointer);
  if(it != PointersMap.end())
      return it->value;
  else
  {
      // your code
  }

File: PolyEngine/Core/Src/SafePtrRoot.cpp [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+	size_t idx = SafePtrRoot::Pointers.GetSize() - 1;
+	SafePtrRoot::PointersMap[pointer] = idx;
+	//return &SafePtrRoot::Pointers[idx];
```
MuniuDev: remove this comment.

PolyEngine/Core/Src/SafePtrRoot.cpp
```diff
@@ -0,0 +1,46 @@
...
+//------------------------------------------------------------------------------
+void SafePtrRoot::ClearPointer(SafePtrRoot *pointer)
+{
+	size_t idx = SafePtrRoot::PointersMap[pointer];
+	SafePtrRoot::Pointers[idx] = nullptr;
+	SafePtrRoot::PointersMap.erase(pointer);
+}
+
+//------------------------------------------------------------------------------
+SafePtrRoot *SafePtrRoot::GetPointer(size_t idx)
```
MuniuDev: Provide const overload.

PolyEngine/UnitTests/Src/SafePtrTests.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <SafePtrRoot.hpp>
+#include <SafePtr.hpp>
+
+using namespace Poly;
+
+TEST_CASE("Safe pointer tests", "[SafePtr]") {
+	SafePtrRoot safePtrObject = SafePtrRoot();
+	SafePtr<SafePtrRoot> safePtr = SafePtr<SafePtrRoot>(&safePtrObject);
+
+	REQUIRE(safePtr->Debug() == safePtrObject.Debug());
```
MuniuDev: You can provide this debug method by implementing new test class that inherits from SafePtrRoot.

PolyEngine/Core/Src/SafePtrRoot.cpp
```diff
@@ -0,0 +1,46 @@
...
+	size_t idx = SafePtrRoot::Pointers.GetSize() - 1;
+	SafePtrRoot::PointersMap[pointer] = idx;
+	//return &SafePtrRoot::Pointers[idx];
+	return idx;
+}
+
+//------------------------------------------------------------------------------
+void SafePtrRoot::ClearPointer(SafePtrRoot *pointer)
+{
+	size_t idx = SafePtrRoot::PointersMap[pointer];
```
MuniuDev: Add assertion for lack of value in PointersMap. Use `getAt` instead of `operator[]`

PolyEngine/Core/Src/SafePtrRoot.cpp
```diff
@@ -0,0 +1,53 @@
...
+void SafePtrRoot::ClearPointer(SafePtrRoot *pointer)
+{
+	size_t idx;
+	try
+	{
+		idx = SafePtrRoot::PointersMap.at(pointer);
+		SafePtrRoot::Pointers[idx] = nullptr;
+		SafePtrRoot::PointersMap.erase(pointer);
+	}
+	catch (std::out_of_range e)
```
MuniuDev: Usage of exceptions is forbidden in performance sensitive code. better use `find` method like this:
```
const auto it = SafePtrRoot::PointersMap.find(pointer);
if(it != SafePtrRoot::PointersMap.end())
{
    SafePtrRoot::Pointers[it->second] = nullptr;
    SafePtrRoot::PointersMap.erase(it);
}
```

PolyEngine/Core/Src/SafePtr.hpp
```diff
@@ -5,64 +5,65 @@
...
 		}
 
-		SafePtr operator=(const SafePtrRoot* pointer)
-		{
-			return SafePtr(pointer);
-		}
+		/// <summary>Alternative way to construct SafePtr instance</summary>
+		/// <param name="pointer">Raw pointer to target object</param>
+		/// <returns>SafePtr instance</returns>
+		SafePtr operator=(const SafePtrRoot* pointer) {	return SafePtr(pointer); }
```
MuniuDev: This operator can be removed alltogether, since we have implicit constructor.

PolyEngine/Core/Src/SafePtr.hpp
```diff
@@ -5,64 +5,65 @@
...
+		bool operator==(const SafePtr<T>& other) const { return this->Get() == other.Get();	}
 
-		bool operator==(const T& other) const
-		{
-			return this->Get() == &other;
-		}
+		/// <summary>Equality operator</summary>
+		/// <param name="other">Raw pointer</param>
+		/// <returns>True, if both SafePtr instance and raw pointer points to the same object. False, if not</returns>
+		bool operator==(const T& other) const {	return this->Get() == &other; }
```
MuniuDev: This should take a  pointer to `T` not its reference, like so:
`bool operator==(const T* other) const { return this->Get() == other; }`
Same for neq operator.

PR #88, opened 2017-11-15
Reset function implementation

Astrael1:


PolyEngine/Engine/Src/Engine.cpp
```diff
@@ -67,9 +67,18 @@ void Engine::RegisterUpdatePhase(const PhaseUpdateFunction& phaseFunction, eUpda
...
+
 //------------------------------------------------------------------------------
 void Engine::Update()
 {
+	if (IsRestartRequested)
+	{
+		Game->Deinit();
+		BaseWorld.reset();
+		Game->Init();
+		BaseWorld = std::make_unique<World>();
```
MuniuDev: BaseWorld should be created before Game init. Have you tested it? It would crash in such case.
Also remeber to set IsRestartRequested to false after handling it.

PR #90, opened 2017-11-17
OpenAL SoundDevice moved to dll.

PiotrMoscicki:
I haven't created proper cmake and macOS files yet but overally everything works (tested only on SGJGame (and for now only SGJGame is adapted for working with new audio)).


MuniuDev:
We are rethinking thse changes under new architecture.

PR #92, opened 2017-11-19
Rendering: Bugfixes

sp0lsh:
### TODO: 
- [x] fix RenderingSandbox missing meshes
- [x] fix SGJ missing meshes
- [x] debug color of SGJ meshes


PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
```diff
@@ -153,3 +152,20 @@ void BlinnPhongRenderingPass::OnRun(World* world, const CameraComponent* camera,
 		glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
 	}
 }
+
+void Poly::BlinnPhongRenderingPass::CreateDummyTexture()
```
MuniuDev: Is this a fallback for meshes without textures? It would be better to make it more distinguishable (e.g. magents/black checkerbox) and move it to mesh loading code. This can be done in separate PR.

sp0lsh: its fallback to white when texture is not present (sometimes on purpose)

PR #97, opened 2017-11-25
Rendering: Spotlights

sp0lsh:


PolyEngine/Engine/Res/Shaders/blinn-phongFrag.shader
```diff
@@ -45,6 +54,8 @@ uniform DirectionalLight uDirectionalLight[8];
 uniform int uDirectionalLightCount;
 uniform PointLight uPointLight[8];
 uniform int uPointLightCount;
+uniform SpotLight uSpotLight[8];
```
MuniuDev: Don't use magic numbers, extract this to define on the top of the file. Like so:
```
#ifndef MAX_SPOTLIGHT_COUNT
#define MAX_SPOTLIGHT_COUNT 8
#endif
```
This way, in the future, we can provide variable count of lights during shader compilation, by appending `#define MAX_SPOTLIGHT_COUNT 12` to the beginning of the shader file. Do it also for the point lights and directional lights. This should also be applied to every constant in shader code.

PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
```diff
@@ -104,6 +118,27 @@ void BlinnPhongRenderingPass::OnRun(World* world, const CameraComponent* camera,
 	}
 	GetProgram().SetUniform("uPointLightCount", pointLightsCount);
 
+	int spotLightsCount = 0;
+	for (auto componentsTuple : world->IterateComponents<SpotLightComponent, TransformComponent>())
```
MuniuDev: `const auto&`

PolyEngine/RenderingDevice/OpenGL/Src/BlinnPhongRenderingPass.cpp
```diff
@@ -104,6 +118,27 @@ void BlinnPhongRenderingPass::OnRun(World* world, const CameraComponent* camera,
...
+		GetProgram().SetUniform(baseName + "Range", spotLightCmp->GetRange());
+		GetProgram().SetUniform(baseName + "CutOff", Cos(1.0_deg * spotLightCmp->GetCutOff()));
+		GetProgram().SetUniform(baseName + "OuterCutOff", Cos(1.0_deg * spotLightCmp->GetOuterCutOff()));
+		GetProgram().SetUniform(baseName + "Position", transformCmp->GetGlobalTranslation());
+		GetProgram().SetUniform(baseName + "Direction", MovementSystem::GetGlobalForward(transformCmp));
+		GetProgram().SetUniform(baseName + "Base.Color", spotLightCmp->GetColor());
+		GetProgram().SetUniform(baseName + "Base.Intensity", spotLightCmp->GetIntensity());
+
+		++spotLightsCount;
+		if (spotLightsCount == MAX_LIGHT_COUNT_SPOT)
```
MuniuDev: We can provide API in `ShaderProgram` that enables seting shader constnts to certain value from C++ code. Sounds good? ;)

sp0lsh: yes

PR #108, opened 2017-12-18
Rendering: Skybox

sp0lsh:


PolyEngine/RenderingDevice/OpenGL/Src/GLRenderingDevice.hpp
```diff
@@ -103,6 +106,7 @@ namespace Poly
 		EnumArray<std::unique_ptr<RenderingPassBase>, ePostprocessRenderPassType> PostprocessRenderingPasses;
 
 		std::unique_ptr<PostprocessQuad> PostprocessRenderingQuad;
+		std::unique_ptr<PrimitiveCube> PrimitiveRenderingCube;
```
MuniuDev: Maybe we could create cube with reversed UV to address fliped textures problem

sp0lsh: Better fix the vertex order in triangles.

sp0lsh:
Cubemaps are difficult to create and iterate. I have prototype of cubemap to lat-long mapping. Will post in another PR.

PolyEngine/RenderingDevice/OpenGL/Src/GLRenderingDevice.hpp
sp0lsh: Vertex order used but it was not very effective.
Scaled UV by vec3(-1.0, 1.0, 1.0) and it was supper effective.
*Skybox flipped*

PR #111, opened 2018-01-02
Project support WiP

MuniuDev:


PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,120 @@
...
+    with fileinput.FileInput(fileToSearch, inplace=True) as file:
+        for line in file:
+            for (tag, val) in tagsAndValues:
+                line = line.replace(tag, val)
+            print(line, end='')
+
+def RunCmake(path, buildDir):
+    # Run cmake update (using undocumented parameters that work for some reason, src: http://cprieto.com/posts/2016/10/cmake-out-of-source-build.html)
+    if os.name == 'nt':
+        os.system(r'cmake -G "Visual Studio 14 2015 Win64" -H' + path + ' -B' + buildDir)
```
DevRayy: Raw string is not necessary here. You should also use .format() to place variables in your string, for example:
`os.system('cmake -G "Visual Studio 14 2015 Win64" -H {} -B {}'.format(path, buildDir))`

Using .format() applies to all string concatenations in this file

PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,120 @@
...
+    print('Creating project', name, 'in', path, 'with engine at', enginePath)
+
+    if os.path.exists(path):
+        raise ValueError('Path', path, 'already exists. Cannot create project there!')
+    
+    path = os.path.abspath(path)
+
+
+    os.makedirs(path)
+    os.makedirs(path + os.sep + 'Build')
```
DevRayy: Consider using something like this when dealing with os.sep:
`os.makedirs(os.sep.join([path, 'Build']))`

PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,120 @@
...
+    ReplaceTagsInFile(path + os.sep + 'CMakeLists.txt', [('$PROJECT_PATH$', name), ('$ENGINE_DIR$', os.path.abspath(enginePath).replace('\\', '/'))])
+    ReplaceTagsInFile(projectPath + os.sep + 'CMakeLists.txt', [('$PROJECT_NAME$', name)])
+
+    RunCmake(path, path + '/Build')
+
+def UpdateProject(path, enginePath):
+    print('Updating project at', path, 'with engine at', enginePath)
+    RunCmake(path, path + '/Build')
+
+#################### SCRIPT START ####################
```
DevRayy: You should use `if __name__ == "__main__":` to indicate where this script actually starts

PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,123 @@
...
+# Functions
+# Create project function
+
+# Setup:
+# $PROJECT_NAME$ - name of the project
+# $PROJECT_PATH$ - relative path of the project (game) directory (relative to root project dir)
+# $GAME_CLASS_NAME$ - name of the game class
+# $ENGINE_DIR$ - absolute path to engine root directory
+
+def ReplaceTagsInFile(fileToSearch, tagsAndValues):
```
DevRayy: According to [PEP8](https://www.python.org/dev/peps/pep-0008/#naming-conventions), function and variable names should be written in snake_case

PolyEngine/Scripts/ProjectTool.py
MuniuDev: I'm using raw string to handle " inside of a string. But good point.

PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,123 @@
...
+
+def UpdateProject(path, enginePath):
+    print('Updating project at', path, 'with engine at', enginePath)
+
+    if not os.path.exists(path):
+        raise Exception('Path', path, 'does not exists. Cannot update project there!')
+
+    RunCmake(path, path + '/Build')
+
+#################### SCRIPT START ####################
```
DevRayy: You should use `if __name__ == "__main__":` to indicate where script starts

PolyEngine/Scripts/ProjectTool.py
```diff
@@ -0,0 +1,123 @@
...
+def CreateProject(name, path, enginePath):
+    print('Creating project', name, 'in', path, 'with engine at', enginePath)
+
+    if os.path.exists(path):
+        raise Exception('Path', path, 'already exists. Cannot create project there!')
+    
+    path = os.path.abspath(path)
+
+    os.makedirs(path)
+    os.makedirs(path + os.sep + 'Build')
```
DevRayy: Consider using this, when dealing with os.sep:
os.makedirs(os.sep.join([path, 'Build']))

PolyEngine/Scripts/ProjectTool.py
DevRayy: You don't have to. If you start a string with single quotes you can use double quotes inside of this string as much as you like and it won't break. Of course it works the other way around too - if you need to use single quotes in your string, just start and end it with double quotes.

What is more, if you want to use ' and " inside your string, you still can do it starting your string with ''' (3 single quotes). In that case even \n won't break your string (this is a pythonic way to make multi-line comments btw).

PR #134, opened 2018-05-08
Fet/Bitmask

OlekJC:
BitMask working with different unsigned data types. In Unit Tests BitList Dynarray size is checked for Debugging purposes.

### DIFF
--- a/PolyEngine/Core/Src/Collections/BitMask.cpp
+++ b/PolyEngine/Core/Src/Collections/BitMask.cpp
@@ -1,0 +1,389 @@
+#include "CorePCH.hpp"
+
+#include <climits>
+
+constexpr u64 ZERO = 0UL;
+
+using namespace Poly;
+
+constexpr BitMask::DataType TYPE_BIT = CHAR_BIT * sizeof(BitMask::DataType);
+
+BitMask::BitMask(size_t size)
+	: BitsNumber(size)
+{
+	size_t arraySize = 0;
+
+	if (size%TYPE_BIT)
+		arraySize = size / TYPE_BIT + 1;
+	else
+		arraySize = size / TYPE_BIT;
+
+	for (int i = 0; i < arraySize; i++)
+		BitList.PushBack(ZERO);
+}
+
+bool BitMask::Reset()
+{
+	BitList.Clear();
+	BitsNumber = 0;
+	return true;
+}
+
+bool BitMask::Toggle(size_t index)
+{
+	HEAVY_ASSERTE(RangeCheck(index), "Out of bounds");
+	BitList[index / TYPE_BIT] ^= 1UL << index%TYPE_BIT;
+	return true;
+}
+
+bool BitMask::operator[](size_t index) const
+{
+	DataType tempChar = BitList[index / TYPE_BIT];
+	tempChar = (tempChar >> index%TYPE_BIT) & 1UL;
+	return tempChar != 0;
+}
+
+bool BitMask::Set(size_t index, bool state)
+{
+	HEAVY_ASSERTE(RangeCheck(index), "Out of bounds");
+
+	size_t bitListIndex = index / TYPE_BIT;
+	size_t bitPosition = index % TYPE_BIT;
+
+	if (state)
+		BitList[bitListIndex] |= (1UL << bitPosition);
+	else
+		BitList[bitListIndex] &= ~(1UL << bitPosition);
+	return true;
+}
+
+bool BitMask::RangeCheck(size_t index)
+{
+	if (index >= BitsNumber || index < 0)
+		return false;
+	return true;
+}
+
+BitMask BitMask::operator|(const BitMask& rhs) const
+{
+	//Equal Dynarray sizes
+	if (BitList.GetSize() == rhs.BitList.GetSize())
+	{
+		size_t tempBitsNumber = 0;
+		if (BitsNumber > rhs.BitsNumber)
+			tempBitsNumber = BitsNumber;
+		else
+			tempBitsNumber = rhs.BitsNumber;
+
+		BitMask temp(tempBitsNumber);
+		for (size_t i = 0; i < BitList.GetSize(); i++)
+			temp.BitList[i] = BitList[i] | rhs.BitList[i];
+		return temp;
+	}
+
+	else if (BitList.GetSize() > rhs.BitList.GetSize())
+	{
+		BitMask temp(BitsNumber);
+		for (size_t i = 0; i < rhs.BitList.GetSize(); i++)
+			temp.BitList[i] = BitList[i] | rhs.BitList[i];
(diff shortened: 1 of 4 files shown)


PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
+#include "CorePCH.hpp"
+
+#include <climits>
+
+constexpr u64 ZERO = 0UL;
+
+using namespace Poly;
+
+constexpr BitMask::DataType TYPE_BIT = CHAR_BIT * sizeof(BitMask::DataType);
```
MuniuDev: `constexpr size_t TYPE_BIT = CHAR_BIT * sizeof(BitMask::DataType);`

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+BitMask::BitMask(size_t size)
+	: BitsNumber(size)
+{
+	size_t arraySize = 0;
+
+	if (size%TYPE_BIT)
+		arraySize = size / TYPE_BIT + 1;
+	else
+		arraySize = size / TYPE_BIT;
```
MuniuDev: Change this whole block to `arraySize = (size + TYPE_BIT - 1) / TYPE_BIT;`

PolyEngine/Core/Src/Collections/BitMask.hpp
```diff
@@ -0,0 +1,47 @@
...
+	public:
+		using DataType = unsigned short;
+
+		BitMask(size_t size = sizeof(DataType));
+
+		//Bitwise operators
+		BitMask operator|(const BitMask& rhs) const;
+		BitMask operator^(const BitMask& rhs) const;
+		BitMask operator&(const BitMask& rhs) const;
+		BitMask& operator~();
```
MuniuDev: This operator should be const and return a copy, just like &, | and ^.

PolyEngine/Core/Src/Collections/BitMask.hpp
```diff
@@ -0,0 +1,47 @@
...
+		BitMask& operator|=(const BitMask& rhs);
+		BitMask& operator^=(const BitMask& rhs);
+		BitMask& operator&=(const BitMask& rhs);
+
+
+		bool operator==(const BitMask rhs) const;
+		bool operator!=(const BitMask rhs) const
+		{
+			return !(*this == rhs);
+		}
```
MuniuDev: Put this whole method in one line. This should accept `const BitMask&` as argument

PolyEngine/Core/Src/Collections/BitMask.hpp
```diff
@@ -0,0 +1,47 @@
...
+		bool operator[](size_t index) const;
```
MuniuDev: Please add @todo annotation: "Convert this method to return bit proxy", or if you want I can tell you how to write this proxy. In general we want to be able to do sth like this: `BitMask b; b[0] = true;` which is not possible right now.

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+			tempBitsNumber = BitsNumber;
+		else
+			tempBitsNumber = rhs.BitsNumber;
+
+		BitMask temp(tempBitsNumber);
+		for (size_t i = 0; i < BitList.GetSize(); i++)
+			temp.BitList[i] = BitList[i] | rhs.BitList[i];
+		return temp;
+	}
+
```
MuniuDev: Remove this blank line

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+		for (size_t i = BitList.GetSize(); i < rhs.BitList.GetSize(); i++)
+			temp.BitList[i] = rhs.BitList[i];
+		return temp;
+	}
+	return BitMask(0);
+}
+
+BitMask BitMask::operator^(const BitMask& rhs) const
+{
+	if (BitList.GetSize() == rhs.BitList.GetSize())
```
MuniuDev: All three operators (&, |, ^) can be written much simpler. Core algorithm logic looks like this:
1. newBitsNumber = max(BitsNumber, rhs.BitsNumber)
2. Perform logic operation on values
3. Trim returned BitMask to size == position of the leftmost bit set to 1.

We can discuss this further via pm.

PolyEngine/Core/Src/Collections/BitMask.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+		//Set,Reset and toggle bits methods
+		bool Set(size_t index, bool state);
+		bool Reset();
+		bool Toggle(size_t index);
+
+		bool Resize(const int offset = 0);
```
MuniuDev: `size_t size` not `const int offset`

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+BitMask& BitMask::operator~()
+{
+	for (auto& x : BitList)
+		x = ~x;
+
+	return *this;
+}
+
+
+bool BitMask::Resize(const int offset)
```
MuniuDev: Resize should apply given size, not current size + offset. That way this method is simple 
`BitsNumber = size; BitList.Resize((size + TYPE_BIT - 1)/TYPE_BIT)`

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+	}
+	return false;
+}
+
+size_t BitMask::BitListIndex(size_t index)
+{
+	return index / TYPE_BIT;
+}
+
+BitMask& BitMask::operator|=(const BitMask& rhs)
```
MuniuDev: No need to write this like that, simply write `{ *this = *this | rhs; }`. Repeat this for other operators overloads.

PolyEngine/Core/Src/Collections/BitMask.cpp
```diff
@@ -0,0 +1,382 @@
...
+		BitsNumber = rhs.BitsNumber;
+		return *this;
+	}
+	return *this;
+}
+
+bool BitMask::operator==(const BitMask rhs) const
+{
+	if (BitsNumber != rhs.BitsNumber)
+		return false;
```
MuniuDev: Not necessarly, if one of the bitmasks is prefixed with a lot of zeroes this would still be true.
It would probably be helpfull to add method called `Trim()` which resizes BitMast to size == position of the leftmost bit set to 1. This way you can first trim, or get size after potential trim operation, and compare them to each other.

PR #138, opened 2018-06-26
Moving world components to root

PiotrMoscicki:


PolyEngine/Engine/Src/ECS/Entity.hpp
```diff
@@ -95,7 +95,7 @@ namespace Poly
 		World* EntityWorld = nullptr;
 
 		std::bitset<MAX_COMPONENTS_COUNT> ComponentPosessionFlags;
-		ComponentBase* Components[MAX_COMPONENTS_COUNT];
+		ComponentBase** Components;
```
MuniuDev: This change introduces memory leak! Let's unify `MAX_COMPONENTS_COUNT` and `MAX_WORLD_COMPONENTS_COUNT` and revert to normal static array of size `MAX_COMPONENTS_COUNT`. This PR shouldn't require any changes in Entity.[ch]pp files.

PolyEngine/Engine/Src/ECS/World.cpp
```diff
@@ -7,42 +7,49 @@ World::World()
...
+			DestroyEntity(child);
+
+		// remove world components
+		for (size_t i = 0; i < MAX_WORLD_COMPONENTS_COUNT; i++)
+			if (RootEntity->Components[i])
+				delete (RootEntity->Components[i]);
+
+		// destroy root
+		RootEntity->~Entity();
+		delete RootEntity.Get();
```
MuniuDev: If we unify WorldComponents and Components (as we should) this change (as in this block of code) is entirely unnecessary, as destruction of world components would be handled by normal ComponentAllocators deletion.

PolyEngine/Engine/Src/ECS/World.cpp
```diff
@@ -7,42 +7,49 @@ World::World()
...
 	return ent;
 }
 
 //------------------------------------------------------------------------------
-Entity * Poly::World::SpawnEntityInternal()
+Entity* Poly::World::SpawnEntityInternal()
 {
 	Entity* ent = EntitiesAllocator.Alloc();
-	::new(ent) Entity(this);
+	::new(ent) Entity(this, RootEntity.Get());
```
MuniuDev: Good catch!

PR #157, opened 2018-09-18
Rendering: Support for #include in shaders

sp0lsh:
Depends on PR with FXAA:
https://github.com/KNTGPolygon/PolyEngine/pull/156

Contains unit tests and fixes to: 
- Array::Insert(size_t idx, T& obj)
- Array::Insert(size_t idx, T&& obj)
- Array::Insert(size_t idx, const Dynarray<T>& arr)
- Array::Insert(size_t idx, const Dynarray<T>&& arr)
- String::GetTrimmed()

Checking if passes CI tests.
Please see and verify **Dynarray<T>& operator=(Dynarray<T>&& rhs)**


PolyEngine/Core/Src/Collections/Dynarray.hpp
```diff
@@ -230,6 +230,43 @@ namespace Poly
...
+		/// <param name="idx">Index in which object should be created.</param>
+		/// <param name="arr">R-value reference to array of objects that should be copied to the container.</param>
+		void Insert(size_t idx, const Dynarray<T>&& arr)
+		{
+			HEAVY_ASSERTE(idx <= GetSize(), "Index out of bounds!");
+			if (Size + arr.GetSize() >= GetCapacity())
+				Reserve(Size + arr.GetSize());
+			MoveDataRight(idx, Size, arr.GetSize());
+			for (size_t i = 0; i < arr.GetSize(); ++i)
+				ObjectLifetimeHelper::CopyCreate(Data + idx + i, std::move(arr[i]));
```
MuniuDev: This should be a call to `MoveCreate`

PolyEngine/Core/Src/Collections/String.cpp
```diff
@@ -7,6 +7,8 @@ using namespace Poly;
 
 const String String::EMPTY = String();
 
+const String String::WHITESPACES = String(" \t\r\n");
```
MuniuDev: Good change, but this should static Dynarray<char> since we don't use it as string.  This also shouldnt be in header at all. Just static field in this cpp.

PolyEngine/Core/Src/Collections/String.hpp
```diff
@@ -195,6 +195,9 @@ namespace Poly {
 		/*CORE_DLLEXPORT*/ friend std::ostream& operator<< (std::ostream& stream, const String& rhs) { return stream << rhs.GetCStr(); }
 
 	private:
+
+		static const String WHITESPACES;
```
MuniuDev: Remove this declaration.

PolyEngine/RenderingDevice/OpenGL/Src/Proxy/GLShaderProgram.cpp
```diff
@@ -9,14 +9,19 @@ UNSILENCE_MSVC_WARNING()
 
 using namespace Poly;
 
+const String GLShaderProgram::SHADERS_INCLUDE_DIR = "Shaders/";
```
MuniuDev: Same deal as with `WHITESPACES`, make it static, remove declaration from header.

PolyEngine/RenderingDevice/OpenGL/Src/Proxy/GLShaderProgram.cpp
```diff
@@ -97,24 +94,68 @@ void GLShaderProgram::Validate()
...
+		
+	Dynarray<String> includedFiles;
+	Dynarray<String> linesToProcess = rawCode.Split("\n");
+	StringBuilder sb;
+	size_t lineCounter = 0;
+	while(lineCounter < linesToProcess.GetSize())
+	{
+		String currLine = linesToProcess[lineCounter].GetTrimmed();
+
+		std::regex includeRegex(R"(^#pragma include\s*\"(\w.*)\")", std::regex::ECMAScript); // regex: ^#pragma include\s*"(\w.*)"
```
MuniuDev: why are we using `#pragma include` and not simply `#include` ?

PolyEngine/RenderingDevice/OpenGL/Src/Proxy/GLShaderProgram.cpp
```diff
@@ -97,24 +94,68 @@ void GLShaderProgram::Validate()
...
+			for (std::cregex_iterator i = regex_begin; i != regex_end; ++i)
+			{
+				const std::cmatch& match = *i;
+
+				String includePath = SHADERS_INCLUDE_DIR + String(match[1].str().c_str());
+				ASSERTE(!includedFiles.Contains(includePath), "LoadShader failed, found recursive includes!");
+				includedFiles.PushBack(includePath);
+					
+				String rawInclude = LoadTextFileRelative(eResourceSource::ENGINE, includePath);
+				linesToProcess.Insert(lineCounter + 1, rawInclude.Split('\n'));
```
MuniuDev: This whole splitting doesn't seem too optimized. Maybe we can put an annotation in comment `// @fixme optimize this` To later revisit this during the refactor. Also the way we handle circular includes here is stragne, why fail if we can simply ommit the include as it's already included?. Just put a warning in console.

PolyEngine/RenderingDevice/OpenGL/Src/Proxy/GLShaderProgram.hpp
```diff
@@ -61,12 +61,16 @@ namespace Poly
 		void RegisterUniform(const String& type, const String& name);
 
 	private:
-		void CompileProgram();
-		void Validate();
+		static const String SHADERS_INCLUDE_DIR;
```
MuniuDev: Remove this declaration.

PR #161, opened 2018-11-07
ProjectTool.py: fix file mode and add a shebang

dos1:


MuniuDev:
Closing due to inactivity from the author. This change will no longer apply due to scripts refactor introduced.

PR #165, opened 2018-12-01
Rendering: Multiple skyboxes

sp0lsh:
multipleskyboxes2


PolyEngine/Engine/Src/Rendering/SkyboxWorldComponent.cpp
```diff
@@ -11,12 +11,27 @@ RTTI_DEFINE_COMPONENT(::Poly::SkyboxWorldComponent)
 SkyboxWorldComponent::SkyboxWorldComponent(const String& panoramaPath, const eResourceSource source)
 	: Tint(Color::WHITE)
 {
-	EquirectPanorama = ResourceManager<TextureResource>::Load(panoramaPath, source, eTextureUsageType::HDR);
+	Init( {panoramaPath}, source);
+}
 
-	gConsole.LogInfo("SkyboxWorldComponent::SkyboxWorldComponent panoramaPath: {}", panoramaPath);
+SkyboxWorldComponent::SkyboxWorldComponent(const Dynarray<String>& panoramaPaths, const eResourceSource source)
```
MuniuDev: shouldn't it be `Dynarra<std::pair<String, eResourceSource>>` instead?

PR #167, opened 2018-12-01
Rendering: Exponential Fog

sp0lsh:
polyengine standalone 01_12_2018 17_11_17

Reference:
https://docs.microsoft.com/en-us/windows/desktop/direct3d9/fog-formulas


PolyEngine/Engine/Src/Rendering/PostprocessSettingsComponent.hpp
```diff
@@ -28,6 +28,9 @@ namespace Poly {
 		float VignetteScale = 0.1f;
 		float AbberationScale = 0.1f;
 
+		Color FogColor = Color::WHITE;
+		float FogDensity = 0.66f;
```
MuniuDev: Shouldn't fog be disabled by default?

PolyEngine/RenderingDevice/OpenGL/Src/TiledForwardRenderer.cpp
```diff
@@ -1312,29 +1312,42 @@ void TiledForwardRenderer::PostGamma(const SceneView& sceneView)
...
+	float planeFar = sceneView.CameraCmp->GetClippingPlaneFar();
+
 	const ScreenSize screenSize = RDI->GetScreenSize();
 
 	float grainScale = 0.01f;
 	float vignetteScale = 0.1f;
 	float abberationScale = 0.1f;
 	Color tint = Color::WHITE;
+	Color fogColor = Color::WHITE;
+	float fogDensity = 0.66f;
```
MuniuDev: Same here. Also why is this duplicated?

PolyEngine/RenderingDevice/OpenGL/Src/TiledForwardRenderer.cpp
```diff
@@ -1312,29 +1312,42 @@ void TiledForwardRenderer::PostGamma(const SceneView& sceneView)
...
 	const PostprocessSettingsComponent* postCmp = sceneView.CameraCmp->GetSibling<PostprocessSettingsComponent>();
 	if (postCmp) {
-		grainScale = postCmp->GrainScale;
-		vignetteScale = postCmp->VignetteScale;
-		abberationScale = postCmp->AbberationScale;
-		gamma = postCmp->Gamma;
-		tint = postCmp->Tint;
+		grainScale		= postCmp->GrainScale;
+		vignetteScale	= postCmp->VignetteScale;
+		abberationScale	= postCmp->AbberationScale;
```
JDuchniewicz: why such indentation?

PolyEngine/RenderingDevice/OpenGL/Src/TiledForwardRenderer.cpp
sp0lsh: These are default values in case when PostProcessSettingsComponent is not present. Should be moved to config soon.

PolyEngine/Engine/Src/Rendering/PostprocessSettingsComponent.hpp
MuniuDev: Good point.

PR #168, opened 2018-12-01
Rendering: SceneView queue sorting

sp0lsh:
SceneView now has geometry in PriorityQueues instead of Dynarray.

Geometry is sorted when filling SceneView. Opauqe is sorted front to back and Translucent is sorted back to front for rendering. During rendering fragment shaders are not called when fragments depth is behind Z-Buffer. If we render huge opaque mesh close camera covering significant part of screen we save execution of fragment shaders that are hidden and would be occluded by huge mesh anyway.

Translucent meshes have to be rendered back to front because of blending.

FINALLY!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # Engine specific files
 console.log
+*.dump # shader content dumps for debuging
 
 # Engine local configurations
 AssetsPathConfig.json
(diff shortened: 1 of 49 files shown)


MuniuDev:
Resolve conflicts and we cane merge this.

PR #169, opened 2018-12-01
Rendering: Particle depth fade

sp0lsh:
Depth fade based on difference in scene depth and fragment depth of particle


MuniuDev:
Resolve conflicts and we cane merge this.

PR #171, opened 2018-12-27
Rtti std vec

rybochodonezzar:
Property for std::vector has common base with Dynarray. eCorePropertyType::DYNARRAY renamed to LIST


PolyEngine/Core/Src/RTTI/RTTIProperty.hpp
```diff
@@ -8,6 +8,7 @@
 #include <Collections/Dynarray.hpp>
 #include <RTTI/CustomTypeTraits.hpp>
 #include <Utils/Logger.hpp>
+#include <vector>
```
MuniuDev: I don't think we need this here, put it in PCH if it's not there.

PR #180, opened 2019-02-03
[WiP] Componentiterator refactor

MuniuDev:


PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+	};
+
+	template<typename PrimaryComponent, typename... SecondaryComponents>
+	class ComponentIterator : public BaseObject<>,
+							public std::iterator<std::forward_iterator_tag, std::tuple<typename std::add_pointer<PrimaryComponent>::type, typename std::add_pointer<SecondaryComponents>::type...>>	
+	{
+		public:
+			explicit ComponentIterator(std::unique_ptr<IEntityIteratorHelper> iter) : Iter(std::move(iter))
+			{
+				UpdateCache();
```
MuniuDev: Just set cache to invalid. It'll update when needed.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+			}
+			bool operator==(const  ComponentIterator& rhs) const { return Iter.get()->get() == rhs.Get()->get(); } //they never  trigger underlying overload, pointer comparison ongoing
+			bool operator!=(const  ComponentIterator& rhs) const { return !(Iter.get()->get() == rhs.Get()->get()); }
+
+			std::tuple<typename std::add_pointer<PrimaryComponent>::type, typename std::add_pointer<SecondaryComponents>::type... > operator*() const //canot be const if  change cache
```
MuniuDev: Better to return const& of the tuple.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+			{
+				return GetCache();
+			}
+			std::tuple<typename std::add_pointer<PrimaryComponent>::type, typename std::add_pointer<SecondaryComponents>::type... > operator->() const
+			{
+				return GetCache();
```
MuniuDev: Same here.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+			}
+
+			ComponentIterator& operator++() { Increment(); return *this; }
+			ComponentIterator operator++(int) { ComponentIterator ret(Iter); Increment(); return ret; } //test for double incrementing etc
+
+			IEntityIteratorHelper* Get() const
```
MuniuDev: Why do we need Helper to be visible outside of the iterator? This method should be private or protected at most. Name of this method requires more consideration.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+			{
+				return Iter.get();
+			}
+		private:
+			void Increment()
+			{
+				//ASSERTE(Iter.get()->isValid(), "Next Iterator is not valid!");	//currently not working as we never set it to nullptr?
+				Iter.get()->increment();
```
MuniuDev: either use `Get()` or `Iter.get()`.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+				//ASSERTE(Iter.get()->isValid(), "Next Iterator is not valid!");	//currently not working as we never set it to nullptr?
+				Iter.get()->increment();	
+				bCacheValid = false;
+			}
+
+			void UpdateCache() const 
+			{
+				Entity* ent = Iter.get()->get();
+				PrimaryComponent* primary = ent->GetComponent<PrimaryComponent>();
+				if (!primary)
```
MuniuDev: Not true. We should assert on primary being nullptr.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+				{
+					bCacheValid = false;
+					return;
+				}
+				if constexpr (!Trait::IsVariadicEmpty<SecondaryComponents...>::value)
+				{
+					Cache = std::make_tuple(primary, primary->template GetSibling<SecondaryComponents>()...);
+					bCacheValid = true;
+					return;
+				}
```
MuniuDev: remove those two lines, only setting cache is necessary here.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+				else
+				{
+					Cache = std::make_tuple(primary);
+				}
+
+				bCacheValid = true;
+			}
+
+			std::tuple<typename std::add_pointer<PrimaryComponent>::type, typename std::add_pointer<SecondaryComponents>::type... > GetCache() const
```
MuniuDev: return const& here. If needed create another version of this method that returns mutable & and is mutable itself.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
```diff
@@ -0,0 +1,109 @@
...
+					UpdateCache();
+				// Entity* ent = Iter.get()->get();
+				// if (!HasComponents<PrimaryComponent, SecondaryComponents...>(ent))
+				// 	ASSERTE(false, "Updated cache does not contain all needed components!"); //@fixme: celeborth, when put in asserte clause it fails (passing to macro fails template deduction?)
+				return Cache;
+			}
+
+			//------------------------------------------------------------------------------
+			template<int zero = 0>
+			bool HasComponents(const Entity* entity) const { return true; }
```
MuniuDev: You never use this method. Do we ensure that the secondary components are never null? I don't think we do. At least not in this piece of code, and it should be handled by the ComponentIterator, not the IteratorHelper.

.

MuniuDev (commented):
ComponentIterator reviewed. I'm not reviewing the IteratorHelper implementation, until we finalize the API.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: When would we need another version that is mutable? For now the method is returning const& and all the members that are modified inside are marked as mutable.

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: Changed to GetIteratorHelper and decided to move this to protected (I think that inheriting classes will have no need of reimplementing this method - unless they want to keep pointers to several IEntityIteratorHelpers)

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/ComponentIterator.hpp
JDuchniewicz: done

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -147,23 +147,23 @@ namespace Poly {
 			return {this};
 		}
 
-		class SceneComponentIteratorHelper : public IEntityIteratorHelper
+		class SceneComponentIteratorPolicy : public IEntityIteratorPolicy
 		{
 			public:
-				SceneComponentIteratorHelper(Entity* entity, std::vector<size_t> required)
+				SceneComponentIteratorPolicy(Entity* entity, size_t primary)
```
MuniuDev: `primary` -> `primaryComponentID`

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -175,32 +175,27 @@ namespace Poly {
...
 				}
-				bool isValid() const override //think where this might be useful for outside world???
+				bool isValid() const override 
 				{
-					return Match != nullptr;
+					return Match != End;
 				}
 			private:
-				void findMatch(Entity* entity, std::vector<size_t> required)
+				void findMatch(Entity* entity, size_t primary)
```
MuniuDev: Fix codding style:
- methods start with capital letters
- there needs to be new line between metods implementations.

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -229,113 +217,22 @@ namespace Poly {
 					Match = &*(scene->GetEntityAllocator().End());
 				}
 				Entity* Match;
-				std::vector<size_t> RequiredComponents;
+				Entity* End;
+				size_t Primary;
```
MuniuDev: `Primary` -> `PrimaryComponentID`

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -229,113 +217,22 @@ namespace Poly {
 					Match = &*(scene->GetEntityAllocator().End());
 				}
 				Entity* Match;
-				std::vector<size_t> RequiredComponents;
+				Entity* End;
```
MuniuDev: Why do we need these values? Why not simply store the iterator? That's why we have the policy abstraction. Also, we don't need to store the `End` we can always extract it from the scene/allocator. This will greatly simplify the code.

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -147,23 +147,23 @@ namespace Poly {
...
 			public:
-				SceneComponentIteratorHelper(Entity* entity, std::vector<size_t> required)
+				SceneComponentIteratorPolicy(Entity* entity, size_t primary)
 				{ 
-					findMatch(entity, required);
-					//do we want to set our required components even in case of invalidated iterator?
+					findMatch(entity, primary);
 				}
-				SceneComponentIteratorHelper(Entity* entity)
+				SceneComponentIteratorPolicy(Entity* entity)
```
MuniuDev: What is this constructor for?

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -147,23 +147,23 @@ namespace Poly {
...
-		class SceneComponentIteratorHelper : public IEntityIteratorHelper
+		class SceneComponentIteratorPolicy : public IEntityIteratorPolicy
 		{
 			public:
-				SceneComponentIteratorHelper(Entity* entity, std::vector<size_t> required)
+				SceneComponentIteratorPolicy(Entity* entity, size_t primary)
 				{ 
-					findMatch(entity, required);
-					//do we want to set our required components even in case of invalidated iterator?
+					findMatch(entity, primary);
```
MuniuDev: Why do we need to call this method? Isn't entity already a match for us?

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -120,11 +120,11 @@ namespace Poly {
 			IteratorProxy(Scene* s) : S(s) {}
 			ComponentIterator<PrimaryComponent, SecondaryComponents...> Begin()
 			{
-				return ComponentIterator<PrimaryComponent, SecondaryComponents...>(S->MakeSceneComponentIteratorHelper<PrimaryComponent, SecondaryComponents...>());
+				return ComponentIterator<PrimaryComponent, SecondaryComponents...>(S->MakeSceneComponentIteratorPolicy<PrimaryComponent, SecondaryComponents...>());
 			}
 			ComponentIterator<PrimaryComponent, SecondaryComponents...> End()
 			{
-				return ComponentIterator<PrimaryComponent, SecondaryComponents...>(S->MakeSceneComponentIteratorHelper<PrimaryComponent, SecondaryComponents...>(true));
+				return ComponentIterator<PrimaryComponent, SecondaryComponents...>(S->MakeSceneComponentIteratorPolicy<PrimaryComponent, SecondaryComponents...>(true));
```
MuniuDev: There should be two separate methods for begin and end, not a bool variable as argument.

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -175,32 +175,27 @@ namespace Poly {
 				{
 					findNextMatch();
 				}
-				bool isValid() const override //think where this might be useful for outside world???
+				bool isValid() const override
```
MuniuDev: Why increment is not simple `++it` ?

PolyEngine/Engine/Src/ECS/Scene.hpp
```diff
@@ -213,14 +208,7 @@ namespace Poly {
...
 						Entity& e = *it;
-						size_t idx = 0;
-						for (auto id : RequiredComponents)
-						{
-							if (!e.GetComponent(id))
-								break;
-							++idx;
-						}
-						if (idx == RequiredComponents.size())
+						if (e.GetComponent(Primary))
```
MuniuDev: Why do we need those find methods? There are completely not needed. This shgould be a simple iterator wrapper.

MuniuDev:
Due to performance constraints we've decided to close this for now and revisit it in the future if needed.

PR #181, opened 2019-02-06
Adding option to build without the QT and the Editor

1stDimension:


MuniuDev:
This needs to be done on dev.

PR #206, opened 2019-07-27
Rtti2

rybochodonezzar:
Commented out exception in UniqueAttribute caused segfault in tests, no idea why.

### DIFF
--- a/PolyEngine/Core/Src/RTTI/RTTI2.hpp
+++ b/PolyEngine/Core/Src/RTTI/RTTI2.hpp
@@ -1,0 +1,208 @@
+#pragma once
+#include <any>
+#include <typeindex>
+#include <vector>
+#include <memory>
+#include <unordered_map>
+#include <algorithm>
+#include <stdexcept>
+
+namespace Poly
+{
+    constexpr auto List = [](auto ...xs) {
+        return [=](auto access) { return access(xs...); };
+    };
+
+    template <typename F, typename X, typename ... XS>
+    void _runList(F f, X x, XS... xs)
+    {
+        f(x);
+        _runList(f, xs...);
+    }
+
+    template <typename F, typename X>
+        void _runList(F f, X x) { f(x); }
+
+    constexpr auto runList = [](auto list, auto f) {
+        list([f](auto... xs){ _runList(f, xs...); });
+    };
+    namespace RTTI2
+    {
+        template <typename T> struct RTTIinfo {
+            constexpr static auto info = List();
+        };
+
+        using AttrMap = std::unordered_map<std::type_index, std::any>;
+
+        template <typename T> struct SimpleAttribute
+        {
+            using container = std::vector<T>;
+            static void add(AttrMap& m, T x)
+            {
+                auto i = m.find(typeid(T));
+                if(i == m.end())
+                    i = m.emplace(typeid(T), container{}).first;
+                std::any_cast<container>(&i->second)->emplace_back(x);
+            }
+        };
+
+        template <typename T> struct DerivedAttribute
+        {
+            using container = std::vector<std::shared_ptr<T>>;
+            template <typename TT> static void add(AttrMap& m, TT x)
+            {
+                static_assert(std::is_base_of<T,TT>::value, "invalid type");
+                auto i = m.find(typeid(T));
+                if(i == m.end())
+                    i = m.emplace(typeid(T), container{}).first;
+				std::any_cast<container>(&i->second)->emplace_back(std::shared_ptr<T>{
+					new TT{ x },
+						[](T* p) { delete static_cast<TT*>(p); }
+				});
+            }
+        };
+
+        template <typename T> struct UniqueAttribute
+        {
+            using container = T;
+            static void add(AttrMap& m, T x)
+            {
+				if (!m.try_emplace(typeid(T), x).second) {}
+                    //throw std::runtime_error("inserting second UniqueAttribute");
+            }
+        };
+
+        template <typename T> struct AttrType {
+            using type = SimpleAttribute<T>;
+        };
+
+		struct classname { const char* name; };
+        template <> struct AttrType<classname> {
+            using type = UniqueAttribute<classname>;
+        };
+
+        struct TypeInfo;
+
+        struct TypeManager {
+            static TypeManager& get()
+            {
+                static TypeManager instance{};
+                return instance;
+            }
+            template <typename T> const TypeInfo& registerOrGetType();
+            std::unordered_map<std::string, std::shared_ptr<TypeInfo>> byName;
+            std::unordered_map<std::type_index, std::shared_ptr<TypeInfo>> byId;
+			TypeManager(const TypeManager& rhs) = delete;
+			TypeManager& operator=(const TypeManager& rhs) = delete;
+            private:
+            TypeManager() = default;
+        };
+
+		namespace { struct baseClassBase {}; }
+		template <> struct AttrType<baseClassBase> { using type = DerivedAttribute<baseClassBase>; };
+        /* this is the only attribute that will not be included in TypeInfo
+         * it is empty anyway; insted it will be used to store pointers to
+         * store base class pointers */
+		template <typename T> struct baseclass : baseClassBase { using type = T; };
+		template <typename T> struct AttrType<baseclass<T>> { using type = DerivedAttribute<baseClassBase>; };
+
+        struct TypeInfo {
+            std::type_index id;
+            std::type_index listId;
+            std::vector<std::reference_wrapper<const TypeInfo>> bases; //maybe weak_ptr
+            AttrMap attrs;
+            std::function<void*()> construct;
+            std::function<void(void*)> destroy;
+            TypeInfo()
+				: id{typeid(void)}
+                , listId{typeid(void)}
+                , bases{}
+                , attrs{}
+                , construct{nullptr}
+                //, serialize{nullptr}
+                {}
+            template <typename T> void init()
+            {
+                runList(RTTIinfo<T>::info, [this](auto x){
+                    using xT = decltype(x);
+                    if constexpr (std::is_base_of<baseClassBase, xT>::value) {
+                        using type = typename xT::type;
+                        bases.emplace_back(TypeManager::get().registerOrGetType<type>());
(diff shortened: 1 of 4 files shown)


PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
+#pragma once
+#include <any>
+#include <typeindex>
+#include <vector>
+#include <memory>
+#include <unordered_map>
+#include <algorithm>
+#include <stdexcept>
+
+namespace Poly
```
MuniuDev: Let's start moving everything in core to new namespaces we want to switch to. In this case it should be:
`pe::core::rtti2` It should also be reflected in directories structure, so place the file in: `PolyEngine/Core/Src/pe/core/rtti2`

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
+#pragma once
+#include <any>
```
MuniuDev: Please include `Defines.hpp` This ensures all project specific stuff is properly configured, eg.: allocators

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+    };
+    namespace RTTI2
+    {
+        template <typename T> struct RTTIinfo {
+            constexpr static auto info = List();
+        };
+
+        using AttrMap = std::unordered_map<std::type_index, std::any>;
+
+        template <typename T> struct SimpleAttribute
```
MuniuDev: All classes/structures should derive from either `Poly::BaseObject<>` for polymorphic classes or `Poly::BaseObjectLiteralType<>`. In second case the inheritance should be `final`. This is very important, as our math structures require alignment of 16, by deriving from those classes we override the new operator making.

I know this class does not hold any data, but this is a matter of principal and codding rules.

PolyEngine/UnitTests/Src/RTTI2Test.cpp
```diff
@@ -0,0 +1,78 @@
...
+};
+template <> struct RTTI2::RTTIinfo<TestClass> {
+	constexpr static auto info = List(testAttr, testAttr, RTTI2::classname{ "testname" });
+};
+
+TEST_CASE("RTTI2 simple attribute", "[RTTI2]") {
+	TestClass tc{};
+	CHECK(tc.typeInfo.id != std::type_index{ typeid(void) });
+	auto ta = tc.typeInfo.get<TestAttr>();
+	CHECK(ta);
```
MuniuDev, line 20: In cases when failed test assertion invalidates further results please use `REQUIRE` variant of the macro. It stops a current test case, which helps us to get rid of false negatives. This happens throughout this file.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+#include <stdexcept>
+
+namespace Poly
+{
+    constexpr auto List = [](auto ...xs) {
+        return [=](auto access) { return access(xs...); };
+    };
+
+    template <typename F, typename X, typename ... XS>
+    void _runList(F f, X x, XS... xs)
```
MuniuDev: For non-public API it would be better to use `namespace impl` instead of underscore

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+                std::any_cast<container>(&i->second)->emplace_back(x);
+            }
+        };
+
+        template <typename T> struct DerivedAttribute
+        {
+            using container = std::vector<std::shared_ptr<T>>;
+            template <typename TT> static void add(AttrMap& m, TT x)
+            {
+                static_assert(std::is_base_of<T,TT>::value, "invalid type");
```
MuniuDev: We have wrappers for assertions that improve their functionality. In case of static assertions those wrappers are only there to improve the consistency. Nevertheless consistency is important. Look up:
* `STATIC_ASSERTE`
* `ASSERTE`
* `HEAVY_ASSERTE`

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+            }
+        };
+
+        template <typename T> struct UniqueAttribute
+        {
+            using container = T;
+            static void add(AttrMap& m, T x)
+            {
+				if (!m.try_emplace(typeid(T), x).second) {}
+                    //throw std::runtime_error("inserting second UniqueAttribute");
```
MuniuDev: Please never commit commented out code.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+                auto i = m.find(typeid(T));
+                if(i == m.end())
+                    i = m.emplace(typeid(T), container{}).first;
+				std::any_cast<container>(&i->second)->emplace_back(std::shared_ptr<T>{
+					new TT{ x },
+						[](T* p) { delete static_cast<TT*>(p); }
+				});
+            }
+        };
+
```
MuniuDev: There are some indentation issues in this area of code. We use tabs.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+        struct TypeManager {
+            static TypeManager& get()
+            {
+                static TypeManager instance{};
+                return instance;
+            }
+            template <typename T> const TypeInfo& registerOrGetType();
+            std::unordered_map<std::string, std::shared_ptr<TypeInfo>> byName;
+            std::unordered_map<std::type_index, std::shared_ptr<TypeInfo>> byId;
```
MuniuDev: These maps could have better names. I suggest `nameToTypeInfo` and `idToTypeInfo`. This makes them more explicit.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+        struct RTTIBase {
+            const TypeInfo& typeInfo;
+			RTTIBase() = delete;
+			RTTIBase(const TypeInfo& ti) : typeInfo(ti) {}
+            template <typename T> bool isOfType() { return isDerivedOrSame<T>(typeInfo); }
+			~RTTIBase() {}
+        };
+		
+		//for some reason this doesn't work, maybe someone will enlighten me why
```
MuniuDev: Template exporting through DLL-s? I'm not sure because I don't know what kind of issue do you have with it.

MuniuDev (changes requested):

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: Those structs are basically traits. I don't remeber any inheritance in our custom type traits. Those will never exist in runtime and i don't see any reason for such ceremony

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: This is the exception that was causing issues in tests. Will fix

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: This looked differently when i was editing it with vim. Damn you Visual Studio!

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: basicly i wanted to avoid ceremony when deriving from RTTIBase, with deriving from this instead, but it kept trying to call default constructor of RTTIBase for some reason

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: Or anonymous namespace perhaps?

MuniuDev: Anonymus namespace will have no effect as this is a header file.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
MuniuDev: In my opinion, it all comes down to consistency. We can, of course, let it slide, but then we have to create a guide for when it's necessary to derive from those classes. Deriving from them always is a lot easier to enforce and cost us nothing both in terms of workload and code optimization. This is my reasoning, but I'm open for the discussion.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
MuniuDev: Then we need to have a better look at it. This behavior is indeed strange. We will discuss this further IRL.

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: As discussed IRL, this doesn't make sense and wont be done
Reason: this doesn't have any practical purpose, and most of those clases behave like type traits which don't exist in runtime and don't derive from BaseObject

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: This couldn't work, deleted

PolyEngine/Core/Src/RTTI/RTTI2.hpp
rybochodonezzar: Can i leave style for later?

MuniuDev: In my experience later is usually never. It's better to wait for a bit than push unfinished change.

MuniuDev (approved):

PR #213, opened 2019-09-28
Replace insert usage with emplace for std::map

JDuchniewicz:
insert creates a temporary copy of the object, while emplace forward passed arguments, minor optimization tweak.

### DIFF
--- a/PolyEngine/Engine/Src/AI/PathfindingSystem.cpp
+++ b/PolyEngine/Engine/Src/AI/PathfindingSystem.cpp
@@ -41,7 +41,7 @@ std::optional<std::vector<::pe::core::math::Vector>> CalculateNewPath(const NavG
 
 	AllNodes.push_back(PathNode(startNode, 0, graph->GetHeuristicCost(startNode, destNode) ));
 	openList.Push(std::make_pair(AllNodes.size() - 1, graph->GetHeuristicCost(startNode, destNode)));
-	minCosts.insert({startNode, 0.f});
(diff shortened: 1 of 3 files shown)


PolyEngine/Engine/Src/AI/PathfindingSystem.cpp
```diff
@@ -41,7 +41,7 @@ std::optional<std::vector<::pe::core::math::Vector>> CalculateNewPath(const NavG
 
 	AllNodes.push_back(PathNode(startNode, 0, graph->GetHeuristicCost(startNode, destNode) ));
 	openList.Push(std::make_pair(AllNodes.size() - 1, graph->GetHeuristicCost(startNode, destNode)));
-	minCosts.insert({startNode, 0.f});
+	minCosts.emplace(std::make_pair(startNode, 0.f));
```
MuniuDev: Correct me if I'm wrong, but this is a NOP change.
What you need to do in such situations is write: `minCosts.emplace(startNode, 0.f);`
This way the object is constructed in place and not via copy construction

MuniuDev (changes requested):

PolyEngine/Engine/Src/AI/PathfindingSystem.cpp
JDuchniewicz: Corrected, emplace calls proper constructor at call site, so std::make_pair was a NOP

PR #215, opened 2019-10-06
Dependency Manager

MuniuDev:


PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <pe/api/deps/DependencyManager.hpp>
+#include <pe/core/utils/Logger.hpp>
+
+using ::pe::core::utils::gConsole;
+
+namespace pe::api::deps {
+
+DependencyManager& DependencyManager::get()
+{
+	static DependencyManager instance;
```
JDuchniewicz, line 12: what's the advantage of static instance compared to static private field in class?

PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	gConsole.LogInfo("[DependencyManager] Unregistered dependency [{}].", dependency->getName());
+
+	resetAll(dependency->getType());
+}
+
+void DependencyManager::registerDependencyPtr(IDependencyPtr* ptr)
+{
+	ASSERTE(ptr, "Cannot register nullptr");
+#if defined(DEBUG)
```
JDuchniewicz, line 56: why check this only in debug mode?

PolyEngine/API/Src/pe/api/deps/DependencyPtr.hpp
```diff
@@ -0,0 +1,72 @@
...
+		return stream << "DependencyPtr[" << ptr.get() << "]";
+	}
+protected:
+	void inject(IDependency* dep) override
+	{
+		ASSERTE(dep, "Cannot inject nullptr!");
+		ASSERTE(dep->getType() == getType(), "Invalid type!");
+		// This static cast is safe, as the integrality of the types in 
+		// dependency injection system is taken care of by the DependencyManager.
+		m_ptr = static_cast<T*>(dep);
```
JDuchniewicz, line 56: Can you explain when such static cast would be unsafe?

PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+	injectAll(dependency);
+}
+
+void DependencyManager::unregisterDependency(IDependency* dependency)
+{
+	ASSERTE(dependency, "Cannot unregister nullptr");
+	auto type = dependency->getType();
+	auto it = m_dependencyMap.find(type);
+	ASSERTE(it != m_dependencyMap.end(), "Cannot unregister non-registered dependency!");
+	ASSERTE(it->second == dependency, "Cannot unregister different dependency!");
```
rybochodonezzar, line 45: Can this even happen?

PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
MuniuDev: This is a Meyers singleton pattern. The main advantage is that the DependencyManager will be lazily constructed when the `get()` is called instead of constructed during static initialization. It may not even be constructed if there is no need for it. It also solves static initialization fiasco.

PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
MuniuDev: It shouldn't if the code is written correctly. Assertions should be used for things that should not ever happen unless there is a bug in the code.

PolyEngine/API/Src/pe/api/deps/DependencyManager.cpp
MuniuDev: This is an expensive O(n) check. I didn't want to include it elsewhere.

PolyEngine/API/Src/pe/api/deps/DependencyPtr.hpp
MuniuDev: There are two basic situations which would cause a failure:
1. dep is not of type T -> This is guaranteed to not be true by the DependencyManager
2. dep virtually derives from DependencyBase -> This is against the rules of using this API and must be enforced during code review.

If you want more info read about down casting.

PR #216, opened 2019-10-19
Implemented IndexedString.

MuniuDev:


MuniuDev:
Should I replace all places in core where we return `const char*` to the safer `CString` ? It basically has no computational overhead.

PR #217, opened 2019-10-22
Add iterators to Queue, change coding style.

JDuchniewicz:
Adding iterator functionality to core::Queue, basing it on std::vector as underlying container

### DIFF
--- a/PolyEngine/Core/Src/pe/core/storage/Queue.hpp
+++ b/PolyEngine/Core/Src/pe/core/storage/Queue.hpp
@@ -80,8 +80,8 @@ namespace pe::core::storage
 			if (getSize() != rhs.getSize())
 				return false;
 
-			for (size_t n = 0; n < getSize(); ++n) // maybe just normal vector comparison for equality suffices?
-				if (m_data[getNthIdx(n)]!= rhs.m_data[rhs.getNthIdx(n)])
+			for (size_t n = 0; n < getSize(); ++n) 
+				if (reinterpret_cast<T*>(&(m_data[getNthIdx(n) * sizeof(T)])) != reinterpret_cast<T*>(&(rhs.m_data[rhs.getNthIdx(n) * sizeof(T)])))
 					return false;
 
 			return true;
@@ -110,7 +110,7 @@ namespace pe::core::storage
 		void clear()
 		{
 			for (size_t i = 0; i < m_size; ++i)
-				::pe::core::memory::ObjectLifetimeHelper::Destroy(m_data + getNthIdx(i));
+				::pe::core::memory::ObjectLifetimeHelper::Destroy(reinterpret_cast<T*>(m_data + sizeof(T) * getNthIdx(i)));
 			m_size = 0;
 			m_head = 0;
 			m_tail = 0;
@@ -122,7 +122,7 @@ namespace pe::core::storage
 		{
 			if (getSize() >= getCapacity())
 				enlarge();
-			::pe::core::memory::ObjectLifetimeHelper::MoveCreate(m_data + m_tail, std::move(obj));
+			::pe::core::memory::ObjectLifetimeHelper::MoveCreate(reinterpret_cast<T*>(m_data + m_tail * sizeof(T)), std::move(obj));
 			advanceIdx(m_tail);
 			m_size++;
 		}
@@ -134,7 +134,7 @@ namespace pe::core::storage
 			if (getSize() >= getCapacity())
 				enlarge();
 			decreaseIdx(m_head);
-			::pe::core::memory::ObjectLifetimeHelper::MoveCreate(m_data + m_head, std::move(obj));
+			::pe::core::memory::ObjectLifetimeHelper::MoveCreate(reinterpret_cast<T*>(m_data + m_head * sizeof(T)), std::move(obj));
 			m_size++;
 		}
 
@@ -156,8 +156,8 @@ namespace pe::core::storage
 		{
 			HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!");
 			decreaseIdx(m_tail);
-			T val = m_data[m_tail];
-			::pe::core::memory::ObjectLifetimeHelper::Destroy(m_data + m_tail);
+			T val = *reinterpret_cast<T*>(&(m_data[m_tail * sizeof(T)]));
+			::pe::core::memory::ObjectLifetimeHelper::Destroy(reinterpret_cast<T*>(m_data + m_tail * sizeof(T)));
 			--m_size;
 			return val;	
 		}
@@ -167,8 +167,8 @@ namespace pe::core::storage
 		T popFront()
 		{
 			HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!");
-			T val = m_data[m_head];
-			::pe::core::memory::ObjectLifetimeHelper::Destroy(m_data + m_head);
+			T val = *reinterpret_cast<T*>(&(m_data[m_head * sizeof(T)]));
+			::pe::core::memory::ObjectLifetimeHelper::Destroy(reinterpret_cast<T*>(m_data + m_head * sizeof(T)));
 			advanceIdx(m_head);
 			--m_size;
 			return val;
@@ -176,19 +176,19 @@ namespace pe::core::storage
 
 		/// <summary>Returns reference to the first element in queue.</summary>
 		/// <returns>Reference to the first element in queue.</returns>
-		T& front() { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return m_data[m_head]; };
+		T& front() { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return *reinterpret_cast<T*>(&(m_data[m_head * sizeof(T)])); };
 
 		/// <summary>Returns reference to the last element in queue.</summary>
 		/// <returns>Reference to the last element in queue.</returns>
-		T& back() { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return m_data[getPrevIdx(m_tail)]; };
+		T& back() { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return *reinterpret_cast<T*>(&(m_data[getPrevIdx(m_tail) * sizeof(T)])); };
 
 		/// <summary>Returns const reference to the first element in queue.</summary>
 		/// <returns>Const reference to the first element in queue.</returns>
-		const T& front() const { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return m_data[m_head]; };
+		const T& front() const { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return *reinterpret_cast<T*>(&(m_data[m_head * sizeof(T)])); };
 
 		/// <summary>Returns const reference to the last element in queue.</summary>
 		/// <returns>Const reference to the last element in queue.</returns>
-		const T& back() const { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return m_data[getPrevIdx(m_tail)]; };
+		const T& back() const { HEAVY_ASSERTE(!isEmpty(), "Trying to access empty queue!"); return *reinterpret_cast<T*>(&(m_data[getPrevIdx(m_tail) * sizeof(T)])); };
 
 		/// <summary>
 		/// Ensures that enough space is available in the queue.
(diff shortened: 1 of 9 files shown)


PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -3,71 +3,66 @@
...
-	template<typename T>
-	class Queue final : public BaseObjectLiteralType<>
-	{
-	public:
-		/// <summary>Base queue constructor that creates empty object with capacity == 0.</summary>
-		Queue() = default;
-
-		/// <summary>Creates queue instance with provided capacity.</summary>
+    template<class T>
+    class Queue : public BaseObject<>
```
MuniuDev: Why change from LiteralType to normal? There is no need for Queue to have vtable.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -3,71 +3,66 @@
...
 	/// </summary>
-	template<typename T>
-	class Queue final : public BaseObjectLiteralType<>
-	{
-	public:
-		/// <summary>Base queue constructor that creates empty object with capacity == 0.</summary>
-		Queue() = default;
-
-		/// <summary>Creates queue instance with provided capacity.</summary>
+    template<class T>
```
MuniuDev: We'd rather use typename instead of class for such cases.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -3,71 +3,66 @@
...
-
+		Queue(Queue<T>&& rhs) { move(std::forward<Queue<T>>(rhs)); }
+		
 		/// <summary>Creates queue instance from initializer list.</summary>
 		/// <param name="list"></param>
-		Queue(const std::initializer_list<T>& list) { PopulateFromInitializerList(list); }
+		Queue(const std::initializer_list<T>& list) { populateFromInitializerList(list); }
 
 		/// <summary>Basic destructor.</summary>
 		~Queue()
```
MuniuDev, line 40: If the destructor is empty, remove it.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
 		}
 
+		template<typename... Args>
+		void emplaceFront(Args&&... args)
+		{
+		}
+	*/
 		/// <summary>Performs removal from the back of the queue.</summary>
-		void PopBack()
+		void popBack()
```
MuniuDev: Let's make pop methods return the popped object. Just remember to move it out of the vector and not to copy.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
 		/// In case there is not enought the queue gets reallocated to new, bigger memory block.
 		/// </summary>
 		/// <param name="capacity">Requested capacity of the queue.</param>
-		void Reserve(size_t capacity)
+		void reserve(size_t capacity)
 		{
-			HEAVY_ASSERTE(capacity >= Capacity, "Capacity too small!");
-			Realloc(capacity);
+			HEAVY_ASSERTE(capacity >= m_capacity, "Capacity too small!");
+			realloc(capacity);
```
MuniuDev, line 202: Why not `m_data.reserve(capacity);` ? Then you can remove `realloc(..)` completely.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
 		}
 
 		/// <summary>Checks whether provided object is present in the queue at least once.</summary>
 		/// <param name="rhs">Searched object.</param>
 		/// <returns>True if present, false otherwise.</returns>
-		bool Contains(const T& rhs) const { return FindIdx(rhs) < GetSize(); }
+		bool contains(const T& rhs) const { return findIdx(rhs) < getSize(); }
```
MuniuDev: Remove all find and contains methods. Let's force user to use `std::algorithm`

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+
+			bool operator==(const Iterator& rhs) const { return idx == rhs.idx; }
+			bool operator!=(const Iterator& rhs) const { return idx != rhs.idx; }
+			Iterator& operator=(const Iterator& rhs) = default;
+
+			T& operator*() const { return ref.m_data.at(idx); }
+
+			Iterator& operator++() { idx = ref.getNextIdx(idx); return *this; }
+			Iterator operator++(int) { Iterator ret(ref, idx); idx = ref.getNextIdx(idx); return ret; }
+			Iterator& operator--() { idx = ref.getNextIdx(idx); return *this; }
```
MuniuDev: Probably should be `getPrevIdx` same for the other operator. Let's make sure we have as much test coverage as possible for this class. We should aim for around 100%.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+			using value_type = T;
+			using difference_type = std::ptrdiff_t;
+			using pointer = T*;
+			using reference = T&;
+
+			bool operator==(const Iterator& rhs) const { return idx == rhs.idx; }
+			bool operator!=(const Iterator& rhs) const { return idx != rhs.idx; }
+			Iterator& operator=(const Iterator& rhs) = default;
+
+			T& operator*() const { return ref.m_data.at(idx); }
```
MuniuDev: What about `operator->` ? Do we get it for free?

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+			realloc(capacity);
 		}
 
 		/// <summary>Checks whether provided object is present in the queue at least once.</summary>
 		/// <param name="rhs">Searched object.</param>
 		/// <returns>True if present, false otherwise.</returns>
-		bool Contains(const T& rhs) const { return FindIdx(rhs) < GetSize(); }
+		bool contains(const T& rhs) const { return findIdx(rhs) < getSize(); }
+		
+		class Iterator : public BaseObject<>
```
MuniuDev: all iterators should be BaseObjectLiteralType, the inheritance should be final.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+			using reference = T&;
+
+			bool operator==(const ConstIterator& rhs) const { return idx == rhs.idx; }
+			bool operator!=(const ConstIterator& rhs) const { return idx != rhs.idx; }
+
+			const T& operator*() const { return ref.m_data.at(idx); }
+
+			ConstIterator& operator++() { idx = ref.getNextIdx(idx); return *this; }
+			ConstIterator operator++(int) { ConstIterator ret(ref, idx); idx = ref.getNextIdx(idx); return ret; }
+			ConstIterator& operator--() { idx = ref.getNextIdx(idx); return *this; }
```
MuniuDev: Change to `getPrevIdx` for both operators --

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+			using reference = T&;
+
+			bool operator==(const ReverseIterator& rhs) const { return idx == rhs.idx; }
+			bool operator!=(const ReverseIterator& rhs) const { return idx != rhs.idx; }
+
+			T& operator*() const { return ref.m_data.at(idx); }
+
+			ReverseIterator& operator++() { increment(); return *this; }
+			ReverseIterator operator++(int) { ReverseIterator ret(ref, idx); increment(); return ret; }
+			ReverseIterator& operator--() { idx = ref.getPrevIdx(idx); return *this; }
```
MuniuDev: Since this is `ReverseIterator` `operator--` should call `getNextIdx`. Same for `ConstReverseIterator`

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+		ConstReverseIterator crend() const { return ConstReverseIterator(*this, m_tail); }
 
 	private:
 		//------------------------------------------------------------------------------
-		size_t GetPrevIdx(size_t idx) const { return (idx == 0) ? Capacity - 1 : idx - 1; }
-		size_t GetNextIdx(size_t idx) const { return (idx + 1) % Capacity; }
-		void DecreaseIdx(size_t& idx) const { idx = GetPrevIdx(idx); }
-		void AdvanceIdx(size_t& idx) const { idx = GetNextIdx(idx); }
-		size_t GetNthIdx(size_t& n)	const { return (Head + n) % Capacity; }
+		size_t getPrevIdx(size_t idx) const { return (idx == 0) ? m_capacity - 1 : idx - 1; }
```
MuniuDev: Just thought of optimization, change this line to `return (m_capacity + idx - 1) % m_capacity;`

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+			ConstReverseIterator operator--(int) { ConstReverseIterator ret(ref, idx); idx = ref.getPrevIdx(idx); return ret; }
+		private:
+			ConstReverseIterator(const Queue<T>& queue, size_t index) : ref(queue), idx(index) {};
+			void increment()
+			{
+				if(idx == ref.m_head)
+					idx = ref.m_tail;
+				else 
+					idx = ref.getPrevIdx(idx);
+			}
```
MuniuDev: The way to implement interators in a circular buffer is too always keep two elements in the queue:
- element before head - which is == queue.rend()
- element after end - which is == queue.end()
I'm not sure whether the iterators work correctly in the current setup. For sure we need good test coverage here.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
-		size_t GetNthIdx(size_t& n)	const { return (Head + n) % Capacity; }
+		size_t getPrevIdx(size_t idx) const { return (idx == 0) ? m_capacity - 1 : idx - 1; }
+		size_t getNextIdx(size_t idx) const { return (idx + 1) % m_capacity; }
+		void decreaseIdx(size_t& idx) const { idx = getPrevIdx(idx); }
+		void advanceIdx(size_t& idx) const { idx = getNextIdx(idx); }
+		size_t getNthIdx(size_t& n)	const { return (m_head + n) % m_capacity; }
 
 		//------------------------------------------------------------------------------
-		size_t FindIdx(const T& rhs) const
+		size_t findIdx(const T& rhs) const
```
MuniuDev: Remove this method.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
+				if (m_data.at(getNthIdx(i)) == rhs)
+					return getNthIdx(i);
 
-			return GetSize();
+			return getSize();
 		}
 
 		//------------------------------------------------------------------------------
-		void Realloc(size_t capacity)
+		void realloc(size_t capacity)
```
MuniuDev, line 345: This method is now broken. What it did before and now doesn't is defragment the queue. Which basically means to copy elements in such a way to make m_head == 0 and m_end == size.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -94,182 +89,310 @@ namespace pe::core::storage
...
-}
+	private:
+		size_t m_head = 0;
+		size_t m_tail = 0;
+		size_t m_size = 0;
+		size_t m_capacity = 0;
+		std::vector<T> m_data;
+    };
+	
+	template<typename T> typename ::pe::core::storage::Queue<T>::Iterator begin(::pe::core::storage::Queue<T>& rhs) { return rhs.begin(); }
```
MuniuDev: Those functions are no longer needed. Remove them.

MuniuDev (changes requested):
Nice overhaul. There are some problems though:
- please fix the indentation. It seems that you're using spaces instead of tabs.
- Please add a lot more unit tests with a lot more tests cases (divide large TEST_CASES, to a lot more smaller ones with one scenario each). You can also try using test sections functionallity in catch.

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
JDuchniewicz: with C++20 typename/class keywords will disapperar, but changing for the time being

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
JDuchniewicz: Currently there is a workaround implemented, please review if it is suitable. It is hidden from outside world, and test do cover it.

JDuchniewicz: proper solution implemented in newest commits

PolyEngine/Core/Src/pe/core/storage/Queue.hpp
```diff
@@ -78,198 +77,337 @@ namespace pe::core::storage
...
+			const size_t physicalSize = m_size + 1;
+			idx = (idx + physicalSize - 1 - m_head) % physicalSize;
+			return idx + m_head;
+		}
+		
+		size_t getNextIdx(size_t idx) const
+		{
+			// PhysicalSize takes guard cell into account
+			const size_t physicalSize = m_size + 1;
+			idx = (idx + 1 - m_head) % physicalSize;
```
MuniuDev: we can still get an overflow here if the head is bigger than idx + 1. Add physicalSize to idx as is done in getPrevIdx.

MuniuDev (changes requested):

MuniuDev (approved):

PR #219, opened 2019-12-01
Add Result utility class

PiotrSquaresMoscicki:
Based on https://doc.rust-lang.org/std/result/enum.Result.html
I'll add comments to this class soon in this pull request.


rybochodonezzar:
Is this final form? i would suggest to add several helper functions here:
(assuming `Result`  = `R` because i'm lazy)
`R<T2,E> bind(R<T1,E> x, R<T2,E>(T1) f)` (monadic bind)
`R<T,E> join(R<R<T,E>,E>)` (monadic join)
variadic versions of rust's `and` and `or` (but i would call them `first` and `last`)
`template <typename Res, typename E, typename ... Xs> auto lift(Res(Xs ... args) f)`
where `auto` is lambda taking all all arguments lifted (ie.  T replaced with R<T,E>) and R<Res,E> with first error or result (perhaps i should do this one?)
also if `lift` is implemented it could make sense to lift all non-assigning operators

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -10,20 +10,39 @@ namespace pe::core::utils
...
+		bool isOk() const { return m_value.index() == 0; }
+		bool isErr() const { return !isOk(); }
+		
+		T& getValue() { return std::get<T>(m_value); }
+		E& getError() { return std::get<E>(m_value); }
+		const T& getValue() const { return std::get<T>(m_value); }
+		const E& getError() const { return std::get<E>(m_value); }
+
+		template <typename U>
+		Result<U, E> and(Result<U, E> res)
```
rybochodonezzar: Why not make those functions standalone and variadic?

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -33,18 +52,67 @@ namespace pe::core::utils
...
+	template <typename E>
+	Result<void, E> join(Result<Result<void, E>, E> res)
+	{
+		if (res.isOk())
+			return Result<void, E>();
+		else
+			return res.getError();
+	}
+
+	using aaa = void(*)();
```
rybochodonezzar: Whaaat is this for?

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -33,18 +52,67 @@ namespace pe::core::utils
...
+
+	template <typename T1, typename T2, typename E>
+	Result<T2, E> bind(Result<T1, E> res, Result<T2, E> (*func)(T1))
```
rybochodonezzar: This takes only function pointer. You casting closure (lambda with capture) to pointer is UB. Use auto

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -10,20 +10,39 @@ namespace pe::core::utils
...
-		T& GetValue() { return std::get<T>(m_value); }
-		E& GetError() { return std::get<E>(m_value); }
-		const T& GetValue() const { return std::get<T>(m_value); }
-		const E& GetError() const { return std::get<E>(m_value); }
-	
+		bool isOk() const { return m_value.index() == 0; }
+		bool isErr() const { return !isOk(); }
+		
+		T& getValue() { return std::get<T>(m_value); }
+		E& getError() { return std::get<E>(m_value); }
```
rybochodonezzar: maybe T& getDefault(T&) (returning default if error) and
T& getHandle(T&(E&)) would be usefull to add?

rybochodonezzar (changes requested):

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: random line i forgot to remove, its already removed

PolyEngine/Core/Src/pe/core/utils/Result.hpp
MuniuDev: I suggest different naming scheme:
`value()`
`error()`
`valueOr(T)`

What `getHandle` is supposed to do?

PolyEngine/Core/Src/pe/core/utils/Result.hpp
MuniuDev: I suggest simply using `operator&` and `operator|`.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
MuniuDev: I think auto in an function argument declaration is C++20 feature. Use std::function.
Also, this as well might be a simple method.

PolyEngine/CMakeListsCommon.cmake
```diff
@@ -86,19 +86,19 @@ else()
...
-    set(Qt5_DIR $ENV{QTDIR})
+    set(Qt5_DIR $ENV{POLY_QTDIR})
 endif()
 
 message("QT DIR = ${Qt5_DIR}")
 
 find_package(Qt5 COMPONENTS Core Widgets)
 if(NOT DEFINED Qt5_FOUND OR NOT Qt5_FOUND)
-	message(FATAL_ERROR "Qt5 not found. When using a standalone Qt5 toolchain, please setup QTDIR environment variable to point to it.")
+	message(FATAL_ERROR "Qt5 not found. When using a standalone Qt5 toolchain, please setup POLY_QTDIR environment variable to point to it.")
```
MuniuDev: This looks like your internal local change. Revert this file.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,117 @@
...
+	Result<T, E> join(Result<Result<T, E>, E> res)
+	{
+		if (res.isOk())
+			return res.getValue();
+		else
+			return res.getError();
+	}
+
+	template <typename E>
+	Result<void, E> join(Result<Result<void, E>, E> res)
```
MuniuDev: Are `join` and `bind` widely used names or are we free to name those functions as we want? I would suggest `collapse` instead of `join`. WDYT?
Also, this as well might be a simple method.

MuniuDev (changes requested):

PolyEngine/Core/Src/pe/core/utils/Result.hpp
JDuchniewicz: FYI there is a std::bind

MuniuDev (changes requested):
I would also like an implicit bool cast to be added.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: like `getDefault`/`valueOr` but uses function taking error and returning default value instead.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: I suggest instead to define for all operators (let's use `+` for example here)
`template <typename X1, typename X2, typename Y, typename E> R<Y,E> operator+(R<X1,E> x1, R<X2,E> x2)`
returning error of `x1` or error of `x2` or `x1 + x2`

then, as mentioned before
`template <typename T, typename E> R<T, E> first(R<T, E> ... xs)`
and overload for ranges
`template <typename T, typename E, template<typename> typename C> R<T,E> first(C<R<T,E>> range)`

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: `join` and `bind` are widely used unless someone is trying to call monads [Warm and fuzzy things](https://www.urbandictionary.com/define.php?term=Warm%20Fuzzy%20Thing) insetead

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: or perhaps ditch `first`/`last`, since those functions discard potentially valuable error information unless `E = void`. Perhaps it would be usefull to separate this case into another class

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: `std::bind` is completly different thing.
I don't think it should be a simple method. Once `c++20` is a thing, it won't be long before someone somewhere (most probably in boost) defines monad concept. We should use it then

PiotrSquaresMoscicki: we can have namespace result for free functions for result

rybochodonezzar:
@MuniuDev agree, but should non-error value be always treated as `true` or should it evaluate to bool value of underlying value? In other words should `(bool)(R(false))` evaluate to `true` or `false`?
I'm on the side of evaluating to `false`, since the other option just duplicates functionality of `isErr()`

PiotrSquaresMoscicki:
@MuniuDev @rybochodonezzar I wouldn't go for any implicit cast because user will wonder what does it evaluate. Similarly to std::optional which doesn't have implicit bool cast and i've never had a problem with it.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
MuniuDev: Also please change T1 function argument to r-value ref and add std::move  for line 112.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		template <typename F>
+		Result<void, F> or (Result<void, F> res)
+		{
+			if (isOk())
+				return Result<void, F>();
+			else
+				return res;
+		}
```
MuniuDev: We need to check the behavior of those functions in Rust, as current implementation is strange.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+	private:
+		std::optional<E> m_value;
+	};
+
+	template <typename T, typename E>
+	Result<T, E> join(Result<Result<T, E>, E> res)
```
MuniuDev: Please test those functions/methods for number of copies/moves they perform upon execution both on T and on E.

MuniuDev (changes requested):

PolyEngine/Core/Src/pe/core/utils/Result.hpp
MuniuDev: What is the use case for this function? I don't see a situation where you can end with a Result with Result inside, as this would mean that you didn't handle the error which you should have handled.

MuniuDev:
Also please add doxygen documentation to all pubic API.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: resolved - first and last were removed

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: done

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: I've removed these functions

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: resolved - i've removed join

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: done

PolyEngine/Core/Src/pe/core/utils/Result.hpp
rybochodonezzar: The idea here was to have this as a part of monadic interface, but true, this is practically less usefull than bind

JDuchniewicz:
Just a general comment, I suggest naming commits in [this](https://chris.beams.io/posts/git-commit/) way, so that we maintain high quality like other repositories :)

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,280 @@
...
+		// @returns true if result stores valid value, false if stores error
+		//		value
+		constexpr bool isOk() const { return m_value.index() == 0; }
+
+		// @returns false if result stores valid value, true if stores error
+		//		value
+		constexpr bool isErr() const { return !isOk(); }
+		
+		// @returns valid value if has valid value, crashes otherwise :(
+		constexpr T& value() { return std::get<T>(m_value); }
```
MuniuDev: Let's get rid of mutable API all together.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+		// @returns valid value if has valid value, crashes otherwise :(
+		constexpr const T& value() const { return std::get<T>(m_value); }
```
MuniuDev: Create change value() to be similar in nature to valueOr() (return T by val, overloaded for l and r-value references).
Add valueRef() which returns const T&.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,280 @@
...
+		constexpr T valueOr(U&& alt) &&
+		{ 
+			if (isOk())
+				return std::move(value());
+			else
+				return static_cast<T>(std::forward<U>(alt));
+		}
+
+		// returns error if this stores an error, crashes otherwise :(
+		constexpr E& error() { return std::get<E>(m_value); }
```
MuniuDev: Do the same as for value().

MuniuDev (changes requested):

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,285 @@
...
+		//		value
+		constexpr bool isOk() const { return m_value.index() == 0; }
+
+		// @returns false if result stores valid value, true if stores error
+		//		value
+		constexpr bool isErr() const { return !isOk(); }
+
+
+
+
```
MuniuDev: Remove extra blank lines

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,285 @@
...
+		// @returns valid value if has valid value, crashes otherwise :(
+		constexpr T& value() & { return std::get<T>(m_value); }
+		constexpr const T& value() const & { return std::get<T>(m_value); }
+		constexpr T&& value() && { return std::get<T>(std::move(m_value)); }
```
MuniuDev, line 173: Return by value not by r-value ref (this disables rvo)

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,285 @@
...
+				return static_cast<T>(std::forward<U>(alt));
+		}
+
+
+
+
+		// returns error if this stores an error, crashes otherwise :(
+		constexpr E& error() & { return std::get<E>(m_value); }
+		constexpr const E& error() const & { return std::get<E>(m_value); }
+		constexpr E&& error() && { return std::get<E>(std::move(m_value)); }
```
MuniuDev, line 215: Return by value not by r-value ref.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+		// @returns true if result stores valid value, false if stores error
+		//		value
+		constexpr bool isOk() const { return m_value.has_value() == false; }
+
+		// @returns false if result stores valid value, true if stores error
+		//		value
+		constexpr bool isErr() const { return !isOk(); }
+
+
```
MuniuDev: Remove extra blank lines.

PolyEngine/Core/Src/pe/core/utils/Result.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+		// returns error if this stores an error, crashes otherwise :(
+		constexpr E& error() & { return m_value.value(); }
+		constexpr const E& error() const & { return m_value.value(); }
+		constexpr E&& error() && { return std::move(m_value).value(); }
```
MuniuDev, line 262: Return by value.

MuniuDev (changes requested):

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: rvo works only for variables which are created within a method and then returned

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: rvo works only for variables which are created within a method and then returned

PolyEngine/Core/Src/pe/core/utils/Result.hpp
PiotrSquaresMoscicki: rvo works only for variables which are created within a method and then returned

MuniuDev (approved):

PR #220, opened 2019-12-01
[WiP] Moving engine subsystems to API library.

MuniuDev:


MuniuDev:
> when do we use hungarian notation vs lowerCamelCase?

Hungarian notation for private/protected members, camelCase for local variables and public memebers.

PR #229, opened 2020-04-15
Add engine build cleaning command.

JDuchniewicz:
Might be useful to expand on available commands, when cmake gets lost and full manual cleaning is needed.


MuniuDev (changes requested):
Rename clean_build_engine_project to clean_engine_project. Create similar script for cleaning game project.

